PR #1, opened 2020-06-17
multi: split commits from loop repo

joostjager:
Pure code extraction from loop repo using `git subtree split`

Loop counterpart: https://github.com/lightninglabs/loop/pull/237


chainnotifier_client.go
```diff
@@ -0,0 +1,239 @@
+package lndclient
+
+import (
+	"context"
+	"fmt"
+	"sync"
+
+	"github.com/btcsuite/btcd/chaincfg/chainhash"
+	"github.com/btcsuite/btcd/wire"
+	"github.com/lightninglabs/loop/swap"
```
guggero, line 10: Can this import be removed? Or do you want to do this in a follow-up PR so this is a pure code move?

joostjager:
I don't know if the tagging is too much, because every merge will probably need a new tag?

guggero:
@carlaKC: The version is weird because the commit you're referencing to isn't in the main repo yet. You have to add a rewrite in faraday's `go.mod` to use Joost's fork.

carlaKC:
> I don't know if the tagging is too much, because every merge will probably need a new tag?

We don't need to tag each merge, because go-mod can do pseudo versions. I think it might be helpful to tag the point where we split out from loop (maybe not) so that other projects using lndclient can just point to that version?

joostjager:
But after the first change to this repo, the need to switch to pseudo version anyway? Btw, after merge it looks like `github.com/lightninglabs/lndclient v0.0.0-20200618064231-15628713b198`

I am fine with tagging (which version to start with?), but don't think it is that useful.

joostjager:
Pushed tag `v1.0.0`.

PR #2, opened 2020-06-18
remove loop dependency

joostjager:
PR #1 left a back-reference to `loop` in place. This is cleaned up in this pr.


network.go
```diff
@@ -0,0 +1,66 @@
+package lndclient
+
+import (
+	"errors"
+
+	"github.com/btcsuite/btcd/chaincfg"
+)
+
+// Network defines the chain that we operate on.
+type Network uint8
```
guggero, line 10: Why not use a string type directly?

tx_utils.go
```diff
@@ -0,0 +1,31 @@
+package lndclient
+
+import (
+	"bytes"
+
+	"github.com/btcsuite/btcd/wire"
+)
+
+// encodeTx encodes a tx to raw bytes.
+func encodeTx(tx *wire.MsgTx) ([]byte, error) {
```
guggero, line 10: Might as well export these as we're building a library anyway.

network.go
```diff
@@ -23,6 +23,26 @@ const (
 	NetworkSimnet
 )
 
+// NetworkFromStr parses a network from string.
+func NetworkFromStr(text string) (Network, error) {
+	switch text {
+	case "mainnet":
+		return NetworkMainnet, nil
+
+	case "testnet":
```
guggero, line 32: Add `testnet3` as well to be compliant with some notations?

lnd_services.go
```diff
@@ -127,25 +126,25 @@ func NewLndServices(cfg *LndServicesConfig) (*GrpcLndServices, error) {
 	macaroonDir := cfg.MacaroonDir
 	if macaroonDir == "" {
 		switch cfg.Network {
-		case "testnet":
+		case NetworkTestnet:
```
guggero, line 129: Couldn't this whole switch statement now be removed as `cfg.Network` is now guarded against being an unknown network?

network.go
joostjager: It isn't the same abstraction, but works too. Changed.

tx_utils.go
joostjager: I think this is functionality to belongs more in `github.com/btcsuite/btcd/wire`

network.go
joostjager: It doesn't match the network returned by `lnd.GetInfo`. Don't really want to add anything we don't need now.

lnd_services.go
joostjager: It isn't guarded any more now, because we move to string

PR #3, opened 2020-06-22
lightningclient: add decode payment request function + extra fields

carlaKC:
Combining a few unrelated changes:
- Add start/end height to ListTransactions
- Add DecodePaymentRequest to lightning client
- Add payment request field to payment struct

Tested this locally with Faraday, which needs the new endpoint.

### DIFF
--- a/lightning_client.go
+++ b/lightning_client.go
@@ -1087,6 +1087,11 @@ type Payment struct {
 	// if the payment is settled.
 	Preimage *lntypes.Preimage
 
+	// PaymentRequest is the payment request for this payment. This value
+	// will not be set for keysend payments and for payments that manually
+	// specify their destination and amount.
+	PaymentRequest string
+
 	// Amount is the amount in millisatoshis of the payment.
 	Amount lnwire.MilliSatoshi
 
@@ -1165,6 +1170,7 @@ func (s *lightningClient) ListPayments(ctx context.Context,
 
 		pmt := Payment{
 			Hash:           hash,
+			PaymentRequest: payment.PaymentRequest,
 			Status:         status,
 			Htlcs:          payment.Htlcs,
 			Amount:         lnwire.MilliSatoshi(payment.ValueMsat),
--- a/lightning_client.go
+++ b/lightning_client.go
@@ -44,8 +44,12 @@ type LightningClient interface {
 	LookupInvoice(ctx context.Context, hash lntypes.Hash) (*Invoice, error)
 
 	// ListTransactions returns all known transactions of the backing lnd
-	// node.
-	ListTransactions(ctx context.Context) ([]Transaction, error)
+	// node. It takes a start and end block height which can be used to
+	// limit the block range that we query over. These values can be left
+	// as zero to include all blocks. To include unconfirmed transactions
+	// in the query, endHeight must be set to -1.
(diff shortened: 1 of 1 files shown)


lightning_client.go
```diff
@@ -1229,3 +1233,70 @@ func (s *lightningClient) ChannelBackups(ctx context.Context) ([]byte, error) {
...
+
+	// Expiry is the time at which the payment request expires.
+	Expiry time.Time
+
+	// Description is a description attached to the payment request.
+	Description string
+
+	// PaymentAddress is the payment address associated with the invoice,
+	// set if the receiver supports mpp.
+	PaymentAddress []byte
```
joostjager, line 1265: This is `[32]byte`

lightning_client.go
```diff
@@ -1087,6 +1087,11 @@ type Payment struct {
 	// if the payment is settled.
 	Preimage *lntypes.Preimage
 
+	// PaymentRequest is the payment request for this payment. This value
+	// will not be set for keysend payments and for payments that manually
+	// specify their destination and amount.
+	PaymentRequest string
```
joostjager, line 1093: Do we need this somewhere? I always had doubts about exposing this even.

carlaKC: I need this in Faraday to detect circular payments. Main logic uses htlc attempts, but legacy payments don't have htlcs, so we're going to fall back to payment request. 

It's not ideal, since some legacy payments won't have pay_req either, but I'd rather cover as much of the edge case as I can.

joostjager: Ok, yes if it is only for legacy it seems no problem

lightning_client.go
```diff
@@ -1229,3 +1233,70 @@ func (s *lightningClient) ChannelBackups(ctx context.Context) ([]byte, error) {
 
 	return resp.MultiChanBackup.MultiChanBackup, nil
 }
+
+// PaymentRequest represents a request for payment from a node.
+type PaymentRequest struct {
+	// Destination is the node that this payment request pays to .
+	Destination string
```
bhandras, line 1246: Optionally we could use `route.Vertex`.

lightning_client.go
```diff
@@ -1229,3 +1233,70 @@ func (s *lightningClient) ChannelBackups(ctx context.Context) ([]byte, error) {
...
+		PaymentAddress: resp.PaymentAddr,
+	}
+
+	// Set our timestamp values if they are non-zero, because unix zero is
+	// different to a zero time struct.
+	if resp.Timestamp != 0 {
+		paymentReq.Timestamp = time.Unix(resp.Timestamp, 0)
+	}
+
+	if resp.Expiry != 0 {
```
bhandras, line 1303: Isn't zero expiry equivalent to one hour?

carlaKC: This should be handled lnd side (ie, we should never get 0 from lnd), and I don't want to start overlaying lnd-level logic here. 

If there's a bug in lnd, I think it's better to have expiry=0 than expiry= Sept 1970

lightning_client.go
carlaKC: nice

lightning_client.go
bhandras: Looked this up quickly, fortunately it is handled on the LND side.

PR #4, opened 2020-07-06
add SignDescriptor struct without unused SigHashes field

joostjager:
Some projects contain code to set this field, while it is actually unused. This PR removes the field.

### DIFF
--- a/signer_client.go
+++ b/signer_client.go
@@ -4,6 +4,7 @@ import (
 	"context"
 
 	"github.com/btcsuite/btcd/btcec"
+	"github.com/btcsuite/btcd/txscript"
 	"github.com/btcsuite/btcd/wire"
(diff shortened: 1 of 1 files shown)


go.sum
```diff
@@ -1,10 +1,14 @@
 cloud.google.com/go v0.26.0/go.mod h1:aQUYkXzVsufM+DwF1aE+0xfcU+56JwCaLick0ClmMTw=
+git.schwanenlied.me/yawning/bsaes.git v0.0.0-20180720073208-c0276d75487e h1:F2x1bq7RaNCIuqYpswggh1+c1JmwdnkHNC9wy1KDip0=
```
wpaulino, line 2: What triggered these changes?

joostjager: I don't know. I reverted them and they didn't come back.

PR #7, opened 2020-08-10
lnd_services: Add option BlockUntilChainSynced to NewLndServices

guggero:
To make it easy for users of the `lndclient` library to specify they want to wait for lnd to be synced to its chain backend before continuing with their client application, the boolean `BlockUntilChainSynced` can now be set to true when creating a new `GrpcLndServices`. The `NewLndServices` function will then block until lnd thinks it's fully synced.

### DIFF
--- a/lnd_services.go
+++ b/lnd_services.go
@@ -21,6 +21,10 @@ import (
 var (
 	rpcTimeout = 30 * time.Second
 
+	// chainSyncPollInterval is the interval in which we poll the GetInfo
+	// call to find out if lnd is fully synced to its chain backend.
+	chainSyncPollInterval = 5 * time.Second
+
 	// minimalCompatibleVersion is the minimum version and build tags
 	// required in lnd to get all functionality implemented in lndclient.
 	// Users can provide their own, specific version if needed. If only a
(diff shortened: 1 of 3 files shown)


lnd_services.go
```diff
@@ -250,6 +260,24 @@ func NewLndServices(cfg *LndServicesConfig) (*GrpcLndServices, error) {
...
 
+	// If requested in the configuration, we now wait for lnd to fully sync
+	// to its chain backend. We do not add any timeout as it would be hard
+	// to determine a sane value. If the initial block download is still in
+	// progress, this could take hours.
+	if cfg.BlockUntilChainSynced {
+		log.Infof("Waiting for lnd to be fully synced to its chain " +
+			"backend, this might take a while")
+
+		err := services.waitForChainSync()
```
carlaKC, line 271: Makes sense that we can't add a timeout, but is there perhaps still a need to allow caller to cancel at this stage? Eg, using this setting in loop, lnd isn't synced yet, we want to kill loopd and come back another day, our usual sigterm won't work. Or am I missing the point entirely with this option?

guggero: No, you're absolutely right. Makes sense that we might want to cancel the wait on shutdown. Added an optional context that can be canceled.

PR #8, opened 2020-08-26
walletkit: expose BumpFee to allow CPFP/RBF

wpaulino:


walletkit_client.go
```diff
@@ -346,3 +353,28 @@ func (m *walletKitClient) ListSweeps(ctx context.Context) ([]string, error) {
...
+
+	_, err := m.client.BumpFee(
+		m.walletKitMac.WithMacaroonAuth(rpcCtx),
+		&walletrpc.BumpFeeRequest{
+			Outpoint: &lnrpc.OutPoint{
+				TxidBytes:   op.Hash[:],
+				OutputIndex: op.Index,
+			},
+			SatPerByte: uint32(feeRate.FeePerKVByte() / 1000),
+			Force:      false,
```
halseth, line 376: Should we allow this to be specified by the caller?

halseth (approved):

walletkit_client.go
wpaulino: We could, but I left it out as it wasn't necessary and seems mostly just used for anchors at the moment.

guggero (approved):
LGTM :+1:

PR #9, opened 2020-08-31
build: bump to lnd v0.11.0-beta

joostjager:
Update pointer to final release of `lnd 0.11.0`. Next step is to tag master so that we have at least one tag that is safe to use.

Suggestions welcome for the tagging name scheme. If we need to add more calls to `lndclient` that already exist in lnd 0.11.0, we need a new tag. It could be something like`lndclient v0.11.0-1.0.0`, but that isn't compatible with our other versioning schemes.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,7 +7,7 @@ require (
 	github.com/btcsuite/btclog v0.0.0-20170628155309-84c8d2346e9f
 	github.com/btcsuite/btcutil v1.0.2
 	github.com/btcsuite/btcwallet/wtxmgr v1.2.0
-	github.com/lightningnetwork/lnd v0.11.0-beta.rc2
+	github.com/lightningnetwork/lnd v0.11.0-beta
 	google.golang.org/grpc v1.24.0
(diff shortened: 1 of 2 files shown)


guggero (approved):
LGTM :+1: 

My suggestion is the following:
`master` always tracks the currently released `lnd` version and features added to `lndclient` that already exist in that version. Tags there could be called `v1.0.0-lnd0.11.0` (to conform to SemVer standards).

We create a new branch for each `lnd` version in development. For example `lnd0.11.1` and `lnd0.12.0`. PRs that add new functionality of those versions must be created against those branches. The branches are regularly rebased against `master` and tagged in the same fashion as `master`: `v1.1.0-lnd0.11.1`.

joostjager:
Discussed offline to tag in format `v1.0.0+lnd0.11.1` (using build metadata)

joostjager:
After more discussion with @guggero we concluded that the scheme above may be overkill for what `lndclient` is. It is probably better to track the lnd version that it supports and use a pre-release version when more existing functionality is added to `lndclient`.

So for example: `v0.11.0-4`

PR #17, opened 2020-10-15
lndclient: adding SendCoins

bhandras:


lightning_client.go
```diff
@@ -2147,3 +2154,31 @@ func (s *lightningClient) Connect(ctx context.Context, peer route.Vertex,
 
 	return err
 }
+
+// SendCoins sends the passed amount of (or all) coins to the passed address.
+// Optionally confirmation target and manual fee in sats per byte can be passed.
```
guggero, line 2159: nit: should mention what the returned value is. And since parameters aren't really optional, should we mention that you can just pass in golang's default value to use the defaults?

guggero (approved):
Nice, LGTM! Just a nit about the comment(s).

lightning_client.go
bhandras: done

PR #19, opened 2020-10-27
lnd 0.11.1: allow custom macaroon to be used instead of requiring all subserver macaroons

guggero:
Addresses the problem described in https://github.com/lightninglabs/loop/issues/299.

This PR adds two functionalities:
1. Allow one custom macaroon to be specified (by providing `CustomMacaroonPath` instead of `MacaroonDir` in `NewLndServices`) instead of needing to have all `lnd` macaroons in the same directory. This allows users to use the `admin.macaroon` or a custom baked one.
2. To make it easy to get a "macaroon recipe" for all required subservers, a new helper function `MacaroonRecipe` is added. This requires the `lnrpc.Lightning.ListPermissions` RPC which was only added in `lnd v0.11.1`.

**NOTE:** This PR is against the `lnd-11-1` branch as it requires a compile-time dependency to that version. This code can also be used with older `lnd` versions but then the `MacaroonRecipe` function will throw an error.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,69 @@
+PKG := github.com/lightninglabs/lndclient
+
+LINT_PKG := github.com/golangci/golangci-lint/cmd/golangci-lint
+
+GO_BIN := ${GOPATH}/bin
+LINT_BIN := $(GO_BIN)/golangci-lint
+
+LINT_COMMIT := v1.18.0
+
+DEPGET := cd /tmp && GO111MODULE=on go get -v
+GOBUILD := GO111MODULE=on go build -v
+GOTEST := GO111MODULE=on go test -v
+
+LINT = $(LINT_BIN) run -v
+GOFILES = $(shell find . -type f -name '*.go')
+
+GREEN := "\\033[0;32m"
+NC := "\\033[0m"
+define print
+	echo $(GREEN)$1$(NC)
+endef
+
+default: build
+
+# ============
+# DEPENDENCIES
+# ============
+$(LINT_BIN):
+	@$(call print, "Fetching linter")
+	$(DEPGET) $(LINT_PKG)@$(LINT_COMMIT)
+
+# ============
+# INSTALLATION
+# ============
+
+build:
+	@$(call print, "Building lndclient.")
+	$(GOBUILD) $(PKG)
+
+# =======
+# TESTING
+# =======
+
+unit:
+	@$(call print, "Running unit tests.")
+	$(GOTEST) ./...
+
+unit-race:
+	@$(call print, "Running unit race tests.")
+	env CGO_ENABLED=1 GORACE="history_size=7 halt_on_errors=1" $(GOTEST) ./...
+
+# =========
+# UTILITIES
+# =========
+
+fmt:
+	@$(call print, "Formatting source.")
+	gofmt -l -w -s $(GOFILES)
+
+lint: $(LINT_BIN)
+	@$(call print, "Linting source.")
+	$(LINT)
+
+.PHONY: default \
+	build \
+	unit \
+	unit-race \
+	fmt \
+	lint
--- a/lightning_client.go
+++ b/lightning_client.go
@@ -154,6 +154,11 @@ type LightningClient interface {
 
 	// NetworkInfo returns stats regarding our view of the network.
 	NetworkInfo(ctx context.Context) (*NetworkInfo, error)
+
+	// ListPermissions returns a list of all RPC method URIs and the
+	// macaroon permissions that are required to access them.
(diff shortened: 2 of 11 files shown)


carlaKC (commented):
Really excited to see this change come in!

Tried it out for Faraday [here](https://github.com/lightninglabs/faraday/pull/88), and there's a little bit of a chicken-and-egg issue in that you need all of the macaroons to get a `lndclient` instance to then call `MacaroonRecipe`. If I'm using it in the way you were thinking we'd use it, could be missing the point?

What do you think about using a basic client rather than the `lightning.Client` interface for `MacaroonRecipe`? That could be a bit smoother for people who are starting up something like Faraday or Loop the first time, just give an admin macaroon and get the permissions needed, bake a new macaroon then run with it? If that wasn't the plan, it's still very useful to have to just update docs!

macaroon_recipes.go
```diff
@@ -0,0 +1,112 @@
...
+		// Get the typed pointer from our map of supported interfaces.
+		ifacePtr, ok := supportedSubservers[pkg]
+		if !ok {
+			return nil, fmt.Errorf("unknown subserver %s", pkg)
+		}
+
+		// From the pointer type we can find out the interface, its name
+		// and what methods it declares.
+		ifaceType := reflect.TypeOf(ifacePtr).Elem()
+		serverName := strings.ReplaceAll(ifaceType.Name(), "Client", "")
```
bhandras, line 64: I like that we can use reflection to gather method names but maybe a simple constant map of rpc -> method name would also suffice (and is easier to check/debug). Is perfectly fine as is, just a comment.

bhandras (commented):
Nice changes @guggero!

lightning_client.go
```diff
@@ -2147,3 +2152,45 @@ func (s *lightningClient) Connect(ctx context.Context, peer route.Vertex,
...
+func (s *lightningClient) ListPermissions(
+	ctx context.Context) (map[string][]MacaroonPermission, error) {
+
+	rpcCtx, cancel := context.WithTimeout(ctx, rpcTimeout)
+	defer cancel()
+
+	rpcCtx = s.adminMac.WithMacaroonAuth(rpcCtx)
+
+	perms, err := s.client.ListPermissions(
+		ctx, &lnrpc.ListPermissionsRequest{},
```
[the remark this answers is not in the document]
guggero: Ah yes, great catch, thanks!

macaroon_recipes_test.go
```diff
@@ -0,0 +1,87 @@
...
+
+func (m *lightningMock) ListPermissions(
+	_ context.Context) (map[string][]lndclient.MacaroonPermission, error) {
+
+	return m.mockPermissions, nil
+}
+
+// TestMacaroonRecipe makes sure the macaroon recipe for all supported packages
+// can be generated.
+func TestMacaroonRecipe(t *testing.T) {
```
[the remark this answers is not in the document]
guggero: Added to GitHub, yes.

macaroon_recipes.go
guggero: If the list of exceptions keeps growing, I agree that we can switch over to a full map instead. But for now, this keeps the code quite small and also doesn't require you to add/change anything, if you add a new method that's called the same as the RPC it calls.

guggero:
Thanks for all the good input, I addressed all comments.

As mentioned, this bumps the requirement for `lnd` to `v0.11.1-beta` and is therefore on its own branch. But after merging this, I'll cherry pick the commits 3b8f3521, e9d23751 and bc071235 to master and create a new tag. That way we can use the easier macaroon usage without needing to update `lnd` necessarily.

bhandras (approved):
LGTM :burrito:

macaroon_recipes.go
carlaKC: > if you add a new method that's called the same as the RPC it calls

nittynit: can see this being forgotten, maybe add a PR template with a reminder about this? I find that pretty useful. Can be in a follow up.

Makefile
```diff
@@ -0,0 +1,75 @@
+PKG := github.com/lightninglabs/lndclient
+ESCPKG := github.com\/lightninglabs\/lndclient
```
carlaKC, line 2: nit: unused?

carlaKC (approved):
LGTM, thanks for getting to this! Going to be really helpful to have 🌹

macaroon_recipes.go
guggero: The unit tests should catch that. But to avoid an extra git push cycle I added it to the checklist :+1:

PR #20, opened 2020-10-30
lightning: add listpeers and address to close channel

carlaKC:


lightning_client.go
```diff
@@ -2192,6 +2195,32 @@ func (s *lightningClient) GetChanInfo(ctx context.Context, channelId uint64) (
 	}, nil
 }
 
+// ListPeers gets a list of pubkeys of the peers we are currently connected to.
+func (s *lightningClient) ListPeers(ctx context.Context) ([]route.Vertex,
```
bhandras, line 2208: Although we don't currently need all the data returned from the ListPeers RPC we ultimately might. It's up to you but maybe best to add everything now?

lightning_client.go
```diff
@@ -1914,10 +1914,18 @@ func (p *ChannelClosedUpdate) CloseTxid() chainhash.Hash {
 // sending an EOF), we close the updates and error channel to signal that there
 // are no more updates to be sent.
 func (s *lightningClient) CloseChannel(ctx context.Context,
-	channel *wire.OutPoint, force bool,
-	confTarget int32) (chan CloseChannelUpdate, chan error, error) {
+	channel *wire.OutPoint, force bool, confTarget int32,
+	addr btcutil.Address) (chan CloseChannelUpdate, chan error,
```
bhandras, line 1918: nit: I'd call `addr` => `deliveryAddr` instead to make it more clear. (And/or extend godoc).

bhandras (approved):
LGTM :100: 
Added two nits to consider.

lightning_client.go
carlaKC: renamed + added godoc :)

lightning_client.go
carlaKC: yeah, likely saves somebody some future pain!

PR #21, opened 2020-11-03
lndclient: add all functions required for lndmon

carlaKC:
As part of github.com/lightninglabs/lndmon/issues/57 , I am going to switchover lndmon to use the full set of lnd servcies (rather than just the basic client), since we will need to add `routerrpc` anyway for htlc stream, we might as well go all the way now. 

This PR adds all the functions/fields that lndmon will require to switchover to the full client. I have a WIP for lndmon [here](https://github.com/carlaKC/lndmon/tree/57-lndlcientswitchover) which is working with this set of changes.


guggero (approved):
LGTM, great additions :100:

PR #23, opened 2020-11-11
router: add subscribe htlc events to router client

carlaKC:
Required for github.com/lightninglabs/lndmon/issues/57

### DIFF
--- a/router_client.go
+++ b/router_client.go
@@ -32,6 +32,11 @@ type RouterClient interface {
 	// payment update stream and an error stream.
 	TrackPayment(ctx context.Context, hash lntypes.Hash) (
 		chan PaymentStatus, chan error, error)
+
+	// SubscribeHtlcEvents subscribes to a stream of htlc events from the
+	// router.
+	SubscribeHtlcEvents(ctx context.Context) (<-chan *routerrpc.HtlcEvent,
+		<-chan error, error)
 }
 
 // PaymentStatus describe the state of a payment.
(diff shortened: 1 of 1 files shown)


router_client.go
```diff
@@ -369,3 +379,239 @@ func marshallHopHint(hint zpay32.HopHint) (*lnrpc.HopHint, error) {
...
+				errChan <- err
+				return
+			}
+
+			// Send the update to into our events channel, or exit
+			// if our context has been cancelled.
+			select {
+			case htlcChan <- event:
+
+			case <-ctx.Done():
```
halseth, line 427: return context error on `errChan`?

router_client.go
```diff
@@ -32,6 +34,14 @@ type RouterClient interface {
 	// payment update stream and an error stream.
 	TrackPayment(ctx context.Context, hash lntypes.Hash) (
 		chan PaymentStatus, chan error, error)
+
+	// SubscribeHtlcEvents subscribes to a stream of htlc events from the
+	// router. In the absence of an shared interface in htlcswitch/
+	// htlcnotfier, we use an empty interface for htlc events. Note that
```
halseth, line 40: hm, would it be better to create such a shared interface/struct for events? Seems like subscribers to this RPC shouldn't have to deal with the internal lnd types.

halseth (changes requested):

router_client.go
carlaKC: Kind of an open question for this repo, whether to go back to lnd's types or create entirely new ones. We are already using quite a few internal types (contract state for invoices, some payment structs), but have also created new ones. I went with the lnd types here because there's so much that would need to be duplicated, 4x event types + 22 different enum values, and then we need to keep track of three things, rather than just lnrpc+lnd.

The empty interface isn't nice though, WDYT about wrapping the lnd types in a set of structs with a common interface for now, and eventually adding the interface in lnd itself?
```
type ForwardingEvent struct{
  htlcSwitch.Forwardingevent
}
``

carlaKC: Discussed offline, just going with protos for now.

halseth (approved):
Looks great 👍

PR #25, opened 2020-11-19
client: add channel acceptor with new fields

carlaKC:
**This PR uses rpc fields which are only exposed in lnd 0.12:** it should not be merged to master until we have cut lnd.


lightning_client.go
```diff
@@ -696,6 +701,128 @@ type NetworkInfo struct {
...
+}
+
+// AcceptorResponse contains the response to a channel acceptor request.
+type AcceptorResponse struct {
+	// Accept indicates whether to accept the channel.
+	Accept bool
+
+	// PendingChanId is the pending channel id to which this response
+	// applies.
+	PendingChanId []byte
```
bhandras, line 790: Do we need this field?

bhandras (approved):
LGTM :1st_place_medal:

lightning_client.go
carlaKC: Could set it directly from the request so that we don't need to set it!

bhandras: Yeah, mostly what I meant is that we pass the request containing `PendingChanId` to the closure and then return it in the response from the same closure. Not super important though. I'm also fine with having it in the response too.

lightning_client.go
```diff
@@ -2703,3 +2830,75 @@ func (s *lightningClient) NetworkInfo(ctx context.Context) (*NetworkInfo,
...
+	}
+
+	errChan := make(chan error, 1)
+
+	s.wg.Add(1)
+	go func() {
+		defer s.wg.Done()
+
+		for {
+			request, err := acceptStream.Recv()
```
guggero, line 2857: Should we select on `ctx.Done()` before this, just in case we get that directly after sending a message. Then we can avoid getting an error on the `Recv()` and quit nicely.

lightning_client.go
```diff
@@ -2703,3 +2830,75 @@ func (s *lightningClient) NetworkInfo(ctx context.Context) (*NetworkInfo,
...
+			if err != nil {
+				errChan <- fmt.Errorf("invalid request sent "+
+					"from lnd: %v", err)
+
+				return
+			}
+
+			resp, err := accept(accReq)
+			if err != nil {
+				errChan <- fmt.Errorf("accept funciton "+
```
guggero, line 2875: nit: typo in function.

lightning_client.go
```diff
@@ -2703,3 +2830,75 @@ func (s *lightningClient) NetworkInfo(ctx context.Context) (*NetworkInfo,
...
+					"failed: %v", err)
+
+				return
+			}
+
+			rpcResp := &lnrpc.ChannelAcceptResponse{
+				Accept:          resp.Accept,
+				PendingChanId:   resp.PendingChanId,
```
guggero, line 2883: We could check if the `PendingChanId` in the response is empty and default to the value in the request. Or just remove it completely as @bhandras suggested. I think the field is only in the response to make sure the request and response are matched, since with streams it's possible to get requests and send responses out of order. And because we do it synchronously here, we could abstract that away from the `lndclient` user.

guggero (approved):
LGTM, very useful to have! Going to need this for the Pool sidecar implementation as well.

One discussion that we might need to have again is: How should we handle merges to the master branch?
Or to be more exact: What `lnd` version compatibility should the master branch reflect? Always the latest released?
Or should the master branch only contain functionality that is backward compatible to at least the last major version and everything that _requires_ an update should live in a branch?

Anyway, to progress with this PR, I created the `lnd-12-0` branch and suggest merging the PR against that for now.

lightning_client.go
carlaKC: Decided to leave it in the request, because I can see some use case for wanting to log the pending channel id in the accept code, and can see us perhaps needing to come back to log it. Removed it from the response because we don't need it there. Sound ok?

lightning_client.go
carlaKC: see comment above!

lightning_client.go
carlaKC: sgtm!

carlaKC:
>  Or should the master branch only contain functionality that is backward compatible to at least the last major version and everything that _requires_ an update should live in a branch?

This approach makes sense to me. Ok to merge anything that has been included in the last tagged major version, and let other stuff live on a branch until we tag the next major version.

> Anyway, to progress with this PR, I created the lnd-12-0 branch and suggest merging the PR against that for now.

Will go ahead with this!

carlaKC:
Added `ctx` with timeout to acceptor function [here](https://github.com/lightninglabs/lndclient/compare/1c83885d966ff95b52747ec9cb8332b6a1d8709b..37d550ba33202695d5e3c5c3c5dcac25478f2231).

lightning_client.go
```diff
@@ -2703,3 +2829,92 @@ func (s *lightningClient) NetworkInfo(ctx context.Context) (*NetworkInfo,
...
+
+				return
+			}
+
+			// Create a child context for our accept function which
+			// will timeout after the timeout period provided.
+			ctxt, cancel := context.WithTimeout(ctx, timeout)
+
+			resp, err := accept(ctxt, accReq)
+			if err != nil {
```
guggero, line 2888: Could cancel directly before this line as nothing async is going on after this anymore (or at least shouldn't). Then we could get rid of the other three calls.

lightning_client.go
```diff
@@ -2703,3 +2829,92 @@ func (s *lightningClient) NetworkInfo(ctx context.Context) (*NetworkInfo,
...
+// acceptor.
+type AcceptorFunction func(context.Context,
+	*AcceptorRequest) (*AcceptorResponse, error)
+
+// AcceptChannel create a channel acceptor using the accept function passed in.
+// The timeout provided will be used to timeout the passed accept closure when
+// it exceeds the amount of time we allow. Note that this amount should be
+// strictly less than lnd's chanacceptor timeout parameter.
+func (s *lightningClient) AcceptChannel(ctx context.Context,
+	timeout time.Duration, accept AcceptorFunction) (chan error, error) {
```
guggero, line 2843: :+1:

go.mod
```diff
@@ -3,11 +3,12 @@ module github.com/lightninglabs/lndclient
...
 
 require (
-	github.com/btcsuite/btcd v0.20.1-beta.0.20200730232343-1db1b6f8217f
+	github.com/btcsuite/btcd v0.20.1-beta.0.20200903105316-61634447e719
 	github.com/btcsuite/btclog v0.0.0-20170628155309-84c8d2346e9f
 	github.com/btcsuite/btcutil v1.0.2
 	github.com/btcsuite/btcwallet/wtxmgr v1.2.0
-	github.com/lightningnetwork/lnd v0.11.0-beta
+	//TODO(carla): bump to 0.12 once cut.
+	github.com/lightningnetwork/lnd v0.11.0-beta.rc4.0.20201118235732-1efef51268d6
```
guggero, line 11: Can we drop this whole commit then, if we're merging against master?

guggero (changes requested):

go.mod
carlaKC: Not sure whether I'm following this, this change has new fields in the chanacceptor response which will only be included in the 0.12 release? 

> Or should the master branch only contain functionality that is backward compatible to at least the last major version and everything that requires an update should live in a branch?

So by this approach, we leave this PR on the 0.12 branch until it's released.

guggero: Ah, I wasn't aware that there were new fields added in `lnd` for this to work. I was confused because you remove the `SubscribeHtlcEvents` code and assumed this made it backward compatible :see_no_evil: 

But I think you need to close the PR and reopen against the 0.12 branch then?

carlaKC:
Replacing this with #27, which is against the 0.12 branch rather than against maser. No code change, just re-create pull.

PR #30, opened 2020-12-04
lndclient: add pubkey to Node

bhandras:
The `DescribeGraph` call returned nodes without the pubkeys. This PR fixes this by extending `lndclient.Node` with a `PubKey` member.


guggero (approved):
LGTM :+1:

PR #31, opened 2020-12-08
lndclient: use Features map instead the deprecated GlobalFeatures in the NodeUpdate

bhandras:
Note: first commit to be updated once LND master is tagged.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,12 +3,12 @@ module github.com/lightninglabs/lndclient
 go 1.13
 
 require (
-	github.com/btcsuite/btcd v0.20.1-beta.0.20200903105316-61634447e719
+	github.com/btcsuite/btcd v0.21.0-beta.0.20201114000516-e9c7a5ac6401
 	github.com/btcsuite/btclog v0.0.0-20170628155309-84c8d2346e9f
 	github.com/btcsuite/btcutil v1.0.2
 	github.com/btcsuite/btcwallet/wtxmgr v1.2.0
 	//TODO(carla): bump to 0.12 once cut.
-	github.com/lightningnetwork/lnd v0.11.0-beta.rc4.0.20201118235732-1efef51268d6
+	github.com/lightningnetwork/lnd v0.11.0-beta.rc4.0.20201208091650-669663bfc22a
 	github.com/stretchr/testify v1.5.1
 	google.golang.org/grpc v1.24.0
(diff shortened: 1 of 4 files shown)


guggero:
Should this be opened against the `lnd-12-0` branch if it requires lnd `v0.12.0`?

We came up with the following plan for handling `lnd` compatibility:
 - master is always backward compatible with the last major version (so currently `lnd v0.11.0`)
 - we create branches for all minor versions and future major versions (there are the branches `lnd-11-1` and `lnd-12-0`) and merge PRs to those branches, if the features **require** that version to work.
 - we rebase the branches if needed and use tags to track versions that we depend on in other projects
 
I'm not sure when exactly we should merge the branches back. Probably we could merge the `lnd-11-1` as soon as `lnd v0.12.0` is out of RC.

@carlaKC is that also how you remember our Keybase convo? Or did I remember something incorrectly?

bhandras:
> Should this be opened against the `lnd-12-0` branch if it requires lnd `v0.12.0`?
> 
> We came up with the following plan for handling `lnd` compatibility:
> 
>     * master is always backward compatible with the last major version (so currently `lnd v0.11.0`)
> 
>     * we create branches for all minor versions and future major versions (there are the branches `lnd-11-1` and `lnd-12-0`) and merge PRs to those branches, if the features **require** that version to work.
> 
>     * we rebase the branches if needed and use tags to track versions that we depend on in other projects
> 
> 
> I'm not sure when exactly we should merge the branches back. Probably we could merge the `lnd-11-1` as soon as `lnd v0.12.0` is out of RC.
> 
> @carlaKC is that also how you remember our Keybase convo? Or did I remember something incorrectly?

SGTM, I'll reopen this against lnd-12-0 just wasn't aware of this.

carlaKC:
> SGTM, I'll reopen this against lnd-12-0 just wasn't aware of this policy.

We should write it down somewhere, since it's only documented in a closed PR rn 😅 

Ran into this the other day and it turns out you don't actually need to close and reopen, you can [switch](https://github.blog/2016-08-15-change-the-base-branch-of-a-pull-request/) base branch.

guggero:
> Ran into this the other day and it turns out you don't actually need to close and reopen, you can switch base branch.

Oooooh, nice! TIL, thank you!

guggero (approved):
LGTM, pending the switch to the `lnd-12-0` branch.

PR #35, opened 2020-12-28
Implement RawTLS and RawMacaroons parameters in LndServicesConfig

jalavosus:
This PR allows an instantiation of LndServicesConfig to use raw macaroon data (passed as bytes or base64-encoded strings) as well as raw TLS certificate data (passed as bytes), allowing for LndServicesConfig to be used programmatically using loaded kubernetes secrets, hardcoded macaroons/certs, etc. 

#### Pull Request Checklist


guggero:
Thanks for the PR!
This is definitely something that can be very useful and I think we should add the feature.
Though I'm sure you have a specific use case in mind, therefore my question before I do a full review:
Would it not be enough to specify _one_ raw macaroon and call it `RawCustomMacaroon`? That would simplify things quite a bit. You'd need to either pass the raw admin macaroon or bake a custom one, but that's where we want to head anyway (no more multiple macaroon files as they're tedious to handle).

raw_macaroons.go
```diff
@@ -0,0 +1,171 @@
...
+	"encoding/hex"
+	"fmt"
+)
+
+// RawMacaroons contains raw []byte macaroons
+// for later hex encoding and use with LND.
+type RawMacaroons struct {
+	AdminMac     []byte
+	InvoiceMac   []byte
+	ChainMac     []byte
```
torkelrogstad, line 14: Name this `ChainNotifier`?

raw_macaroons.go
```diff
@@ -0,0 +1,171 @@
...
+	SignerMac    []byte
+	WalletKitMac []byte
+	RouterMac    []byte
+	ReadOnlyMac  []byte
+}
```
torkelrogstad, line 19: I think it'd be better if these struct fields didn't "stutter". The current names contain "macaroon" twice, and I think it's fine to omit this from the field names. We'd then end up with `RawMacaroons.Admin` ++, which I think is nicer

raw_macaroons.go
```diff
@@ -0,0 +1,171 @@
...
+
+// Returns a RawMacaroons object populated with raw []byte macaroons.
+// Best used if loading macaroons directly from files, or if reading from
+// Kubernetes secrets containing raw byte data (for example)
+func NewRawMacaroons(adminMac []byte, invoiceMac []byte,
+	chainMac []byte, signerMac []byte, walletKitMac []byte,
+	routerMac []byte, readOnlyMac []byte) *RawMacaroons {
```
torkelrogstad, line 26: I think this method can be removed. There's already quite few methods exposed for creating a new struct, and I think it's a good thing to expose as few public methods as possible from a package. Could possibly remove some of the other public methods as well?

torkelrogstad (commented):
Tested that I can connect to LND by passing in `[]byte` values.

jalavosus:
> Would it not be enough to specify _one_ raw macaroon and call it `RawCustomMacaroon`? That would simplify things quite a bit. 

I actually like this idea a lot, and would make certain use cases much easier to implement!

jalavosus:
@guggero and @torkelrogstad, I just pushed a slight refactor to this PR -- instead of using a self-rolled `RawMacaroon` type, `LndServicesConfig` takes a `CustomMacaroon` parameter, which is the raw `[]byte` slice of either `admin.macaroon` or a self-baked macaroon with all necessary client permissions.

guggero:
Concept ACK, thanks for the changes!

The PR needs a rebase. It would also be great if you could clean up the commit structure and split the changes into small and logical commits. There are also a few style nits that should be addressed.
Please see https://github.com/lightningnetwork/lnd/blob/master/docs/code_contribution_guidelines.md as we try to adhere to those rules in all our projects.

lnd_services.go
```diff
@@ -150,43 +162,21 @@ func NewLndServices(cfg *LndServicesConfig) (*GrpcLndServices, error) {
...
-			)
+	var (
+		macaroonDir   string
+		loadMacDirErr error
+	)
 
-		default:
-			return nil, fmt.Errorf("unsupported network: %v",
-				cfg.Network)
+	if cfg.CustomMacaroon == nil {
```
halseth, line 176: style nit: reverse this check, check non-nil path instead

lnd_services.go
```diff
@@ -210,12 +200,21 @@ func NewLndServices(cfg *LndServicesConfig) (*GrpcLndServices, error) {
...
-	readonlyMac, err := loadMacaroon(
-		macaroonDir, defaultReadonlyFilename, cfg.CustomMacaroonPath,
-	)
-	if err != nil {
-		return nil, err
+	var readonlyMac serializedMacaroon
+	if cfg.CustomMacaroon == nil {
+		var loadMacErr error
+
+		readonlyMac, loadMacErr = loadMacaroon(
```
halseth, line 207: should move this into the above if

halseth (approved):
Looks pretty good, needs a squash and rebase :)

PR #36, opened 2021-01-05
client: optionally wait for lnd to be unlocked

carlaKC:
This PR adds an option to wait for lnd's wallet to be unlocked before lndclient returns, similar to what we have done waiting for the chain to sync. 

#### Pull Request Checklist

### DIFF
--- a/lightning_client.go
+++ b/lightning_client.go
@@ -878,6 +878,10 @@ func (s *lightningClient) GetInfo(ctx context.Context) (*Info, error) {
 		return nil, err
 	}
 
+	return newInfo(resp)
+}
+
+func newInfo(resp *lnrpc.GetInfoResponse) (*Info, error) {
 	pubKey, err := hex.DecodeString(resp.IdentityPubkey)
 	if err != nil {
 		return nil, err
--- a/lnd_services.go
+++ b/lnd_services.go
@@ -11,6 +11,7 @@ import (
 	"github.com/btcsuite/btcd/chaincfg"
 	"github.com/btcsuite/btcutil"
 	"github.com/lightningnetwork/lnd/lncfg"
+	"github.com/lightningnetwork/lnd/lnrpc"
 	"github.com/lightningnetwork/lnd/lnrpc/verrpc"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
@@ -24,6 +25,10 @@ var (
 	// chainSyncPollInterval is the interval in which we poll the GetInfo
 	// call to find out if lnd is fully synced to its chain backend.
 	chainSyncPollInterval = 5 * time.Second
+
+	// defaultUnlockedInterval is the default amount of time we wait between
+	// checks that the wallet is unlocked.
+	defaultUnlockedInterval = 5 * time.Second
 
 	// minimalCompatibleVersion is the minimum version and build tags
 	// required in lnd to get all functionality implemented in lndclient.
@@ -97,13 +102,23 @@ type LndServicesConfig struct {
 	// block download is still in progress.
 	BlockUntilChainSynced bool
 
+	// BlockUntilUnlocked denotes that the NewLndServices function should
+	// block until lnd is unlocked.
+	BlockUntilUnlocked bool
+
+	// UnlockInterval sets the interval at which we will query lnd to
+	// determine whether lnd is unlocked when BlockUntilUnlocked is true.
+	// This value is optional, and will be replaced with a default if it is
+	// zero.
+	UnlockInterval time.Duration
+
 	// CallerCtx is an optional context that can be passed if the caller
 	// would like to be able to cancel the long waits involved in starting
 	// up the client, such as waiting for chain sync to complete when
-	// BlockUntilChainSynced is set to true. If a context is passed in and
-	// its Done() channel sends a message, the wait for chain sync is
-	// aborted. This allows a client to still be shut down properly if lnd
-	// takes a long time to sync.
+	// BlockUntilChainSynced is set to true, or waiting for lnd to be
+	// unlocked when BlockUntilUnlocked is set to true. If a context is
+	// passed in and its Done() channel sends a message, these waits will
+	// be aborted. This allows a client to still be shut down properly.
 	CallerCtx context.Context
 }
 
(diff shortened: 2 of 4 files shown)


lnd_services.go
```diff
@@ -97,12 +101,23 @@ type LndServicesConfig struct {
...
-	// takes a long time to sync.
-	ChainSyncCtx context.Context
+	// BlockUntilUnlocked denotes that the NewLndServices function should
+	// block until lnd is unlocked.
+	BlockUntilUnlocked bool
+
+	// UnlockInterval sets the interval at which we will query lnd to
+	// determine whether lnd is unlocked when BlockUntilUnlocked is true.
+	// This value is optional.
+	UnlockInterval *time.Duration
```
guggero, line 111: nit: use 0 as the default instead of a pointer?

lightning_client.go
```diff
@@ -837,6 +837,44 @@ func newLightningClient(conn *grpc.ClientConn,
...
+// unlocked.
+func (u *unlockerClient) unlocked(ctx context.Context) bool {
+	// Try to unlock lnd with no password. We expect to receive an error
+	// in all cases:
+	// wallet locked: password is incorrect
+	// wallet unlocked: unlocker not active
+	// no password: unlocker not active
+	// However, we check for a nil error to prevent a panic, we check for
+	// the nil case anyway, and log if we do not fail as expected.
+	_, err := u.client.UnlockWallet(ctx, &lnrpc.UnlockWalletRequest{})
```
guggero, line 860: This is a great idea!
Unfortunately this seems to cause some problems:
 - This will try to unlock the wallet once per second. Because of the scrypt operation, this itself takes multiple seconds and a lot of RAM.
 - There seems to be a deadlock/race when doing the actual unlock. The user call tries to open the wallet DB at the same time this method does. Eventually `lnd` gives up and logs this error: `[ERR] LNWL: Failed to open database: timeout`

I think this can quite easily be fixed by reversing the call: Try to call `GetInfo` and return `true` if `codes.Unimplemented` is no longer returned.

guggero (commented):
Nice idea and very useful!
Tested it and ran into an issue but that should be easy to fix.

lnd_services.go
```diff
@@ -374,11 +399,94 @@ func (s *GrpcLndServices) waitForChainSync(ctx context.Context) error {
...
+		}
+
+		// If we do not get a rpc error code, something else is wrong
+		// with the call, so we fail.
+		rpcErrorCode, ok := status.FromError(err)
+		if !ok {
+			return nil, err
+		}
+
+		// If the main rpc server is unimplemented, we need to fall
```
guggero, line 447: nit: move case specific comment above `case` keyword.

lnd_services.go
```diff
@@ -25,6 +25,10 @@ var (
 	// call to find out if lnd is fully synced to its chain backend.
 	chainSyncPollInterval = 5 * time.Second
 
+	// defaultUnlockedInterval is the default amount of time we wait between
+	// checks that the wallet is unlocked.
+	defaultUnlockedInterval = 2 * time.Second
```
guggero, line 30: I've tried this twice on my machine. Once it worked and once it didn't (was doing some compilation work so the machine was slow), throwing a "connection refused" error. I think we should increase this to 5 seconds, just to give slower machines a bigger chance. The slightly longer delay until the client is ready should be worth not having to start the client again (for example LiT crashing with a connection refused error).

lnd_services.go
```diff
@@ -374,11 +399,94 @@ func (s *GrpcLndServices) waitForChainSync(ctx context.Context) error {
...
+	}
+
+	if waitForUnlocked {
+		log.Info("Waiting for lnd to unlock")
+	}
+
+	for {
+		info, err := client.GetInfo(ctx)
+		if err == nil {
+			return info, nil
```
guggero, line 431: Nice to directly re-use the info in the success case!

guggero (approved):
Code looks good to me, nice work :tada: 
Just two small nits.

lnd_services.go
carlaKC: Could also allow more than one conn refused after unlock, say 3? That way faster machines can exit faster?
Def agreed that it's worth the wait, the 2s is just a magic nr.

guggero: Sure, retrying multiple times would also work, if you are okay with rewriting the logic again :sweat_smile:

carlaKC:
Updated this to use the [wait for ready](https://github.com/grpc/grpc/blob/master/doc/wait-for-ready.md) grpc call option to deal with the case where lnd is unlocking. Behavior is now as follows:
- lnd down: will wait for `waitInterval` then fail
- lnd locked: will query lnd periodically, backing off `waitInterval` between queries, until lnd is unlocked or ctx cancelled
- lnd unlocked, race: will wait `waitInterval` for the main server to come up after unlock

guggero (approved):
tACK! Nice solution with `WaitForReady`, LGTM :100:

lnd_services.go
```diff
@@ -374,11 +399,90 @@ func (s *GrpcLndServices) waitForChainSync(ctx context.Context) error {
...
+		}
+
+		// At this point, we know lnd is locked, so we wait for our
+		// interval, exiting if context is cancelled.
+		select {
+		case <-ctx.Done():
+			return nil, ctx.Err()
+
+		case <-time.After(waitInterval):
+
```
bhandras, line 476: nit: extra nl

lnd_services_test.go
```diff
@@ -156,3 +162,110 @@ func TestLndVersionCheckComparison(t *testing.T) {
...
+		},
+		{
+			name:     "do not wait for unlock",
+			errors:   []error{unlockErr},
+			expected: unlockErr,
+		},
+		{
+			name:         "wait for unlock",
+			waitUnlocked: true,
+			errors:       []error{unlockErr, nil},
```
bhandras, line 239: nit: could add a test where `unlockErr` is returned more than once.

lnd_services_test.go
```diff
@@ -156,3 +162,110 @@ func TestLndVersionCheckComparison(t *testing.T) {
...
+		test := test
+
+		t.Run(test.name, func(t *testing.T) {
+			mock := newLockLndMock(test.errors)
+
+			_, err := getLndInfo(
+				test.context, mock, "readonlymac",
+				test.waitUnlocked, 0,
+			)
+			require.Equal(t, test.expected, err)
```
bhandras, line 268: nit: maybe you could also check that `mock.callCount == len(test.errors)`

bhandras (approved):
With the latest change it's much cleaner, nice work @carlaKC!

lnd_services_test.go
carlaKC: done!

lnd_services_test.go
carlaKC: nice, done 👍

PR #37, opened 2021-01-10
Implement macaroon-permission-based client loading

jalavosus:
This commit changes core behavior of LndServices; instead of relying on
macaroons for all rpc clients to be available, it instead checks the
permissions of loaded macaroon files, and determines which clients can
be loaded from there.

If a read-only macaroon is not available for client compatibility
checking, or admin permissions are not available for a Lightning client,
the initialization function will return an error.

#### Pull Request Checklist

### DIFF
--- a/chainnotifier_client.go
+++ b/chainnotifier_client.go
@@ -3,6 +3,7 @@ package lndclient
 import (
 	"context"
 	"fmt"
+	"gopkg.in/macaroon-bakery.v2/bakery"
 	"sync"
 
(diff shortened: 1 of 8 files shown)


jalavosus:
This PR would work really nicely with the behavior implemented in #35, and would allow for more selective client usage.

jalavosus:
I also acknowledge that this implementation is kinda messy, and any suggestions for cleaning it up are very much welcome.

PR #38, opened 2021-01-11
lndclient: guard against hidden decode error when GlobalFeatures is nil

bhandras:
      in `lnd_services.go` are updated. (Not needed)


guggero (approved):
LGTM :tada:

PR #42, opened 2021-02-22
lndclient: ChanInfo to include pending htlc values and direction

bhandras:
#### Pull Request Checklist

~- [x] Version compatibility matrix in the README and minimal required version
      in `lnd_services.go` are updated.~

### DIFF
--- a/lightning_client.go
+++ b/lightning_client.go
@@ -276,6 +276,9 @@ type ChannelInfo struct {
 	// NumPendingHtlcs is the count of pending htlcs on this channel.
 	NumPendingHtlcs int
 
+	// PendingHtlcs stores the pending HTLCs (amount and direction).
+	PendingHtlcs []PendingHtlc
+
 	// CSVDelay is the csv delay for our funds.
 	CSVDelay uint64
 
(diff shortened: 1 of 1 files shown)


lightning_client.go
```diff
@@ -1310,6 +1323,15 @@ type InvoiceHtlc struct {
 	ResolveTime time.Time
 }
 
+// PendingHtlc represents a HTLC that is currently pending on some channel.
+type PendingHtlc struct {
+	// Incoming is true if the HTLC is incoming.
+	Incoming bool
+
+	// Amount is the amount in millisatoshis that this HTLC represents.
+	Amount btcutil.Amount
```
guggero, line 1332: Should use `lnwire.MilliSatoshi` if it's in fact millisatoshis.

guggero (approved):
LGTM, just one comment about the HTLC amount :tada:

lightning_client.go
bhandras: It's in satoshis though

guggero: Okay, then fixing the comment should be enough.

bhandras: Sorry the comment was copy pasted. Fixed now.

PR #45, opened 2021-03-08
mod: bump to final version of lnd v0.12.0

guggero:
This PR bumps the `lnd-12-0` branch to the final version of `lnd`.

#### Pull Request Checklist

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,8 +7,7 @@ require (
 	github.com/btcsuite/btclog v0.0.0-20170628155309-84c8d2346e9f
 	github.com/btcsuite/btcutil v1.0.2
 	github.com/btcsuite/btcwallet/wtxmgr v1.2.0
-	//TODO(carla): bump to 0.12 once cut.
-	github.com/lightningnetwork/lnd v0.12.0-beta.rc6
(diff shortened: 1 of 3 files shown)


carlaKC (approved):
🕺

carlaKC:
When we merge the 0.12 branch into master, are we going to squash it down at all? Could be nice to have this commit first and get rid of all the intermediate mod bumps. Not sure what we did last time.

guggero:
Yeah, we can do that, sure. We haven't yet merged any of the branches. Had a discussion about it recently with @bhandras and we decided to not yet merge but instead continue with the rebasing pattern.
Because once we merge, we can't add new functionality to our other projects without also increasing the minimum backward compatible version.

What's your opinion on this, @carlaKC? How many versions back should we try to keep backward compatible?

carlaKC:
> instead continue with the rebasing pattern

Yeah that sounds good to me! There's still a lot of older APIs missing in `lndclient` so I'm definitely for allowing space to add those without bumping minimum version.

> How many versions back should we try to keep backward compatible?

This is a magic number, but maybe 2? We could also just see how the rebase burden goes and play it by ear.

bhandras (approved):

PR #49, opened 2021-05-06
multi: add custom rpc timeout to lndclient

carlaKC:
Add the option to overwrite our existing hard-set 30 second timeout for lndclient rpc calls.

There is an argument for making this a per-call option, but I think that it's reasonable to have the option to be able to override the default value for all calls? Adding this now doesn't exclude the option of coming back to add per-call timeouts if we want them. 

#### Pull Request Checklist

### DIFF
--- a/chainnotifier_client.go
+++ b/chainnotifier_client.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"sync"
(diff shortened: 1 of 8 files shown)


guggero (approved):
Very nice, LGTM :100:

PR #55, opened 2021-05-19
lnd_services: make unlock wait compatible with lnd 0.13

guggero:
The new version 0.13 of lnd doesn't return the same RPC error codes
anymore if its wallet is still locked. We need to do some string
matching here to avoid needing to pull in the 0.13 version as a
dependency which would nullify our efforts to have backward compatible
branches of lndclient.

#### Pull Request Checklist

### DIFF
--- a/lnd_services.go
+++ b/lnd_services.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"net"
 	"path/filepath"
+	"strings"
 	"time"
 
 	"github.com/btcsuite/btcd/chaincfg"
@@ -64,6 +65,16 @@ var (
 	// required for lndclient to work.
 	DefaultBuildTags = []string{
 		"signrpc", "walletrpc", "chainrpc", "invoicesrpc",
+	}
+
+	// lnd13UnlockErrors is the list of errors that lnd 0.13 and later
+	// returns when the wallet is locked or not ready yet.
+	lnd13UnlockErrors = []string{
+		"waiting to start, RPC services not available",
+		"wallet not created, create one to enable full RPC access",
+		"wallet locked, unlock it to enable full RPC access",
+		"the RPC server is in the process of starting up, but not " +
+			"yet ready to accept calls",
 	}
 )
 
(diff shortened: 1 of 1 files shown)


carlaKC (approved):
LGTM! 

Just to confirm, the idea is to add this here so that wait for unlocked works for old + 0.13 clients, and then we will get rid of this on the 0.13 branch and just use the unlocker? 

Tested the following cases with `loopd` / v0.12 + v0.13 and all in order!
- Lnd not running at all at `loopd` startup, client shuts down
- Lnd locked at `loopd` startup, client waits for unlock
- Lnd unlocked at `loopd` startup, client ok

guggero:
> Just to confirm, the idea is to add this here so that wait for unlocked works for old + 0.13 clients, and then we will get rid of this on the 0.13 branch and just use the unlocker?

Yes, the `lnd-13-0` branch already has its own unlocking logic that uses the new status RPC. When rebasing that branch I'm going to remove this logic again as we require `lnd v0.13.0` to run that branch anyway.

lnd_services.go
```diff
@@ -448,6 +459,14 @@ func getLndInfo(ctx context.Context, ln lnrpc.LightningClient,
 		// Unlocked, ok: lnd is unlocked, no error
 		// Unlocked, not ok: lnd is unlocked but in bad state, err
 		//
+		// For nodes of version 0.13 and later, we don't get any of the
+		// Unimplemented/Unavailable codes anymore since the RPC server
+		// is up the whole time. Instead we need to look at the error
+		// message itself. We can't just use the state RPC in that case
+		// either because that would mean we'd need to pull in the 0.13
```
halseth, line 466: Good comment, now I understand the problem better 👍

halseth (approved):
Great, turned out not so bad 💯

guggero:
Thanks for the reviews!
I exported the `IsUnlockError()` function so we can also use it in LiT. Did a smoke test and this seems to work fine in the combinations I tested.
Going to merge and rebase the other branches.

PR #56, opened 2021-05-25
lndclient: add (incomplete) QueryRoutes

bhandras:
#### Pull Request Checklist

~- [ ] Version compatibility matrix in the README and minimal required version
      in `lnd_services.go` are updated.~
~- [ ] Update `macaroon_recipes.go` if your PR adds a new method that is called
  differently than the RPC method it invokes.~

### DIFF
--- a/lightning_client.go
+++ b/lightning_client.go
@@ -886,12 +886,6 @@ type QueryRoutesResponse struct {
 }
 
 var (
-	// ErrMalformedServerResponse is returned when the swap and/or prepay
-	// invoice is malformed.
-	ErrMalformedServerResponse = errors.New(
-		"one or more invoices are malformed",
-	)
-
 	// ErrNoRouteFound is returned if we can't find a path with the passed
 	// parameters.
 	ErrNoRouteFound = errors.New("no route found")
--- a/lightning_client.go
+++ b/lightning_client.go
@@ -168,6 +168,11 @@ type LightningClient interface {
 	// macaroon permissions that are required to access them.
 	ListPermissions(ctx context.Context) (map[string][]MacaroonPermission,
 		error)
+
+	// QueryRoutes can query LND to return a route (with fees) between two
+	// vertices.
+	QueryRoutes(ctx context.Context, req QueryRoutesRequest) (
+		*QueryRoutesResponse, error)
 }
 
 // Info contains info about the connected lnd node.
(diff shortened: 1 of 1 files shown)


lightning_client.go
```diff
@@ -806,6 +811,80 @@ type NetworkInfo struct {
...
 
+// QueryRoutesRequest is the request of a QueryRoutes call.
+type QueryRoutesRequest struct {
+	// Source is the optional source vertex of the route.
+	Source *route.Vertex
+
+	// PubKey is the destination vertex.
+	PubKey route.Vertex
+
+	// LastHop is the optional last hop befor the destination.
```
carlaKC, line 822: s/befor/before

lightning_client.go
```diff
@@ -806,6 +811,80 @@ type NetworkInfo struct {
...
+	LastHop *route.Vertex
+
+	// RouteHints represents the different routing hints that can be used to
+	// assist the router. These hints will act as optional intermediate hops
+	// along the route.
+	RouteHints [][]zpay32.HopHint
```
carlaKC, line 828: What does "optional intermediate hops" mean? I think we only supply hints for the final hop? 
Nitty though, I guess in theory we could have hints for other private channels that don't belong to the last hop?

lightning_client.go
```diff
@@ -3098,3 +3177,88 @@ func (s *lightningClient) ListPermissions(
...
+		return nil, err
+	}
+
+	resp, err := s.client.QueryRoutes(rpcCtx, rpcReq)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(resp.Routes) == 0 {
+		return nil, fmt.Errorf("no route found")
```
carlaKC, line 3246: Can we define this error so that callers don't need to string match?

lightning_client.go
```diff
@@ -3098,3 +3177,88 @@ func (s *lightningClient) ListPermissions(
...
+	}
+
+	route := resp.Routes[0]
```
carlaKC, line 3249: How does lnd sort routes if there are more than one? Might make sense to always return the cheapest route (perhaps 0 index is already that?)

carlaKC (commented):
Looking good! Why is this "incomplete", because we only return one route?

lightning_client.go
bhandras: Hop hints are used to simply extend the graph and while the router knows about them, inclusion in the final route is not guaranteed.

lightning_client.go
bhandras: LND always returns one route, the `repeated` property in the proto is just there for historical reasons as far as I can see.

bhandras:
> Looking good! Why is this "incomplete", because we only return one route?

Incomplete because a bunch of things are not returned (TLV, MPP records, etc).

lightning_client.go
bhandras: done

lightning_client.go
bhandras: "Repurposed" an error for this as it seems there were some unused one in the file.

guggero (approved):
LGTM :tada:

lightning_client.go
carlaKC: Ah I see, TIL.

carlaKC (approved):
🌟

PR #58, opened 2021-06-07
invoices: handle EOF from subscribe single invoice

carlaKC:
lightningnetwork/lnd/pull/5207 terminates our single invoice stream once a final state has been delivered to the stream, which results in an `EOF` being sent to the client. This PR updates `SubscribeSingleInvoice` to handle the EOF is the same way that we do for `TrackPayment`, closing the update channel to indicate that we're done sending updates.

Although this functionality is not in lnd 0.11 (the version on our master branch of lndclient), I've opened this against master so that master properly works with 0.13.

#### Pull Request Checklist

### DIFF
--- a/invoices_client.go
+++ b/invoices_client.go
@@ -3,6 +3,7 @@ package lndclient
 import (
 	"context"
 	"errors"
+	"io"
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


yyforyongyu (approved):
LGTM👍

Issue #63, opened 2021-08-18
Value rounded to sat in AddInvoice


sbres:
In the call AddInvoice, the parameter Value that is in mSat is rounded to a full sat in order to be sent in the Value field of the Invoice.
https://github.com/lightninglabs/lndclient/blob/157df0838abcd9fc2212aeadc6948bd9f05fa956/lightning_client.go#L1191
This creates a silent error because if you try to create an invoice of less than a Sat, an invoice for 0 sat is created. This creating an invoice that does not reflect what was requested.

I see that the Invoice structure has the field `ValueMsat`, I think this would be a better fit, the loss of precision will be avoided and the parameters will match the invoice.

guggero:
Fixed by https://github.com/lightninglabs/lndclient/pull/64.

PR #64, opened 2021-08-18
lndclient: Use field ValueMsat in invoice to avoid rounding

sbres:
Change AddInvoice method to avoid invoice rounding to sat by using the Invoice field ValueMsat instead.


sbres:
Related to issue #63

guggero (approved):
Good catch, LGTM :tada:

carlaKC (approved):
Thanks for the PR!

PR #68, opened 2021-09-02
router: add htlcinterceptor

carlaKC:
Htlc interceptor has been in lnd since 0.11, so we can add this to master. 

#### Pull Request Checklist

### DIFF
--- a/lnd_services.go
+++ b/lnd_services.go
@@ -324,6 +324,9 @@ func NewLndServices(cfg *LndServicesConfig) (*GrpcLndServices, error) {
 		log.Debugf("Wait for invoices to finish")
 		invoicesClient.WaitForFinished()
 
+		log.Debugf("Wait for router to finish")
+		routerClient.WaitForFinished()
+
 		log.Debugf("Lnd services finished")
 	}
 
--- a/router_client.go
+++ b/router_client.go
@@ -3,8 +3,10 @@ package lndclient
 import (
 	"context"
 	"crypto/rand"
+	"errors"
 	"fmt"
 	"io"
+	"sync"
 	"time"
 
 	"github.com/btcsuite/btcutil"
@@ -20,6 +22,10 @@ import (
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 )
+
+// ErrRouterShuttingDown is returned when a long-lived call is killed because
+// the router is shutting down.
+var ErrRouterShuttingDown = errors.New("router shutting down")
 
 // RouterClient exposes payment functionality.
 type RouterClient interface {
@@ -42,6 +48,14 @@ type RouterClient interface {
 	// router.
 	SubscribeHtlcEvents(ctx context.Context) (<-chan *routerrpc.HtlcEvent,
 		<-chan error, error)
+
+	// InterceptHtlcs intercepts htlcs, using the handling function provided
+	// to respond to htlcs. This function blocks, and can be terminated by
+	// canceling the context provided. The handler provided should exit on
+	// context cancel, and must be thread-safe. On exit, all htlcs that are
+	// currently held will be released by lnd.
+	InterceptHtlcs(ctx context.Context,
+		handler HtlcInterceptHandler) error
 }
 
 // PaymentStatus describe the state of a payment.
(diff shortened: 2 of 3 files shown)


router_client.go
```diff
@@ -247,11 +262,76 @@ type SendPaymentRequest struct {
...
+	// AmountInMsat is the incoming htlc amount.
+	AmountInMsat lnwire.MilliSatoshi
+
+	// AmountOutMsat is the outgoing htlc amount.
+	AmountOutMsat lnwire.MilliSatoshi
+
+	// IncomingExpiryHeight is the expiry height of the incoming htlc.
+	IncomingExpiryHeight uint32
+
+	// OutgoingExpiry is the expiry height of the outgoing htlcs.
```
bhandras, line 284: nit: godoc is a bit off `s/OutGoingExpiry/OutgoingExpiryHeight`

lnd_services.go
```diff
@@ -324,6 +324,10 @@ func NewLndServices(cfg *LndServicesConfig) (*GrpcLndServices, error) {
 		log.Debugf("Wait for invoices to finish")
 		invoicesClient.WaitForFinished()
 
+		// TODO[carla] stop (w/ quit channel close) vs waitforfinished here?
```
bhandras, line 327: nit: TODO is not actually a TODO anymore? (Since afaict `stop()` does request stop and also waits.)

router_client.go
```diff
@@ -560,3 +646,203 @@ func (r *routerClient) SubscribeHtlcEvents(ctx context.Context) (
...
+		}
+	}
+
+	// Start a goroutine that consumes interception requests from lnd and
+	// sends them into our requests channel for handling. The requests
+	// channel is not buffered because we expect all requests to be handled
+	// until this function exits, at which point we expect our context to
+	// be canceled or quit channel to be closed.
+	requestChan := make(chan InterceptedHtlc)
+	r.wg.Add(1)
```
bhandras, line 698: There's a (very small) chanche that someone calls `InterceptHtlcs` while the `LndServices` cleanup function is called. In this case there's an even smaller chance that we intercept an htlc but we quit right after that (selecting on the `quit` channel) never sending the htlc to the `handler`.  A possible fix could be to always send intercepted htlcs to the `requestChan` and then before shutting down any non handled htlcs should be released inside the other for loop (or maybe in the `stop` function). wdyt?

carlaKC: I was a bit unsure of the internals of the interceptor when I wrote this (hence all the scattered todos). 

Spent some time looking into it, and it seems that the [interceptor](https://github.com/lightningnetwork/lnd/blob/master/lnrpc/routerrpc/forward_interceptor.go#L69) will always [release](https://github.com/lightningnetwork/lnd/blob/master/lnrpc/routerrpc/forward_interceptor.go#L210) all our currently held htlcs, so I think it's ok if we don't pass htlcs on to the handler in our shutdown case? 

Will update the comments in this PR to reflect this, requested review before coming back and updating w/ my new understanding of the interceptor, sorry!

lnd_services.go
carlaKC: ahyep, removing this. Renamed `stop` to `WaitForFinished` so that it's more consistent, I _think_ that's what I'm trying to say here as well.

router_client.go
bhandras: Indeed that's reassuring! Thanks for pointing to the relevant part.

bhandras (approved):
LGTM :tada:

router_client.go
```diff
@@ -560,3 +649,209 @@ func (r *routerClient) SubscribeHtlcEvents(ctx context.Context) (
...
+// InterceptHtlcs intercepts htlcs on a node, using the handler function
+// provided to provide the interceptor with interception decisions. The handler
+// provided may block, but must exit if the context passed in is canceled. It
+// is also responsible for releasing htlcs on exit.
+//
+// There are a few ways in which this method can exit:
+// - ctx canceled: the calling client cancels
+// - r.quit: the router is shut down
+// - lnd stream error: lnd has exited
+// - handler error: a critical error occurred while handing a htlc
```
yyforyongyu, line 662: 👍

router_client.go
```diff
@@ -560,3 +649,209 @@ func (r *routerClient) SubscribeHtlcEvents(ctx context.Context) (
...
+	// be canceled or quit channel to be closed.
+	requestChan := make(chan InterceptedHtlc)
+	r.wg.Add(1)
+	go func() {
+		defer r.wg.Done()
+
+		for {
+			// Do a quick check whether our client context has been
+			// canceled so that we can exit sooner if needed.
+			if ctx.Err() != nil {
```
yyforyongyu, line 708: If the client context has been canceled, shouldn't the below stream return an error?

router_client.go
```diff
@@ -560,3 +649,209 @@ func (r *routerClient) SubscribeHtlcEvents(ctx context.Context) (
...
+			}
+
+			hash, err := lntypes.MakeHash(request.PaymentHash)
+			if err != nil {
+				sendErr(err)
+				return
+			}
+
+			if request.IncomingCircuitKey == nil {
+				sendErr(errors.New("incoming circuit key " +
```
yyforyongyu, line 725: nit: could also make it a global variable.

yyforyongyu (approved):
LGTM👍

router_client.go
carlaKC: I _think_ that it only returns the error once the server tries to send a new item (was testing this locally and noticed that it sometimes took a little time to exit when this ctx was canceled). Could be wrong there though. Adding this just speeds us up a little. Will do dig around the grpc code, because I'm also curious about it.

router_client.go
carlaKC: Decided against it because this is really unexpected behaviour from lnd, just added it to prevent a panic. Don't expect to need to match on this error.

carlaKC:
Just updated some comments in last push - [diff](https://github.com/lightninglabs/lndclient/compare/3e10ff1715a5bc6fe1d8a4992e81f4ad3d8e6d56..2a0275c9c4e087282b03b90318f8fba089805743).

PR #70, opened 2021-09-21
Make lndclient compatible with upcoming lnd v0.14.0-beta release

guggero:
I've just pushed the `lnd-14-0` branch which before this PR is equal to the `lnd-13-0` branch. This PR updates that initial state to be compatible with the upcoming `lnd v0.14.0-beta` release (currently identified by `v0.13.99-beta`).

#### Pull Request Checklist


state_client.go
```diff
@@ -174,6 +181,9 @@ func unmarshalWalletState(rpcState lnrpc.WalletState) (WalletState, error) {
 	case lnrpc.WalletState_RPC_ACTIVE:
 		return WalletStateRPCActive, nil
 
+	case lnrpc.WalletState_SERVER_ACTIVE:
```
Roasbeef, line 184: Perhaps we should just ignore all unknown states here? As otherwise, we need to update this each time we add a new state.

yyforyongyu (approved):
Thanks for picking this up!

lightning_client.go
```diff
@@ -3564,3 +3574,86 @@ func (s *lightningClient) CheckMacaroonPermissions(ctx context.Context,
...
+	timeout time.Duration, intercept InterceptFunction) (chan error, error) {
+
+	interceptStream, err := s.client.RegisterRPCMiddleware(
+		s.adminMac.WithMacaroonAuth(ctx),
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	// We are expected to send an initial registration message immediately.
```
carlaKC, line 3599: Very nice handling this on behalf of the caller!

lightning_client.go
```diff
@@ -3564,3 +3574,86 @@ func (s *lightningClient) CheckMacaroonPermissions(ctx context.Context,
...
+
+	errChan := make(chan error, 1)
+
+	s.wg.Add(1)
+	go func() {
+		defer s.wg.Done()
+
+		for {
+			select {
+			case <-ctx.Done():
```
carlaKC, line 3621: Are we ok to only exit on caller context here, or do we need a quit channel? 

Asking because as is `WaitForFinished()` will not exit until the caller's context for `RegisterRPCMiddleware` is cancelled, even if we're shutting down lnd services.

With the current way we use `LndServices`, we always cancel our context for all calls before we call `Close`, but wondering whether it's apparent to an outside user that you must do so for this to exit?

carlaKC (approved):
lgtm, just one question re shutting down middleware interceptor.

lightning_client.go
guggero: Hmm, good point. I've always looked at the `lndclient` instance as being more or less stateless so would expect the context needing to be canceled. Also I just copied this straight from the `ChannelAcceptor` code which behaves the same way.
Maybe we need to document this more clearly?
Looking at any of the other subscription methods, they all also only quit on context cancel.

state_client.go
guggero: I thought about this a bit and don't think it's a good idea. Because those new states likely mean something has changed in the business logic. And we also possibly need to implement some backward compatible logic here anyway. For example I needed to add the `ReadyForGetInfo()` method to make sure this works with both 0.13.x and 0.14.x nodes. So I think it's good to error out on unknown states.

lightning_client.go
carlaKC: Yeah was just wondering about it, since we're just using cancel everywhere as is seems fine to continue with that pattern. Perhaps we can just drop a comment on the `Stop` method for lnd services that you have to cancel caller ctx to shutdown?

PR #72, opened 2021-10-13
lndclient: add all boolean params for listChannels

arshbot:
#### Pull Request Checklist


carlaKC:
Closing due to inactivity. Re-open if we need this.

PR #81, opened 2021-11-23
Add Mission Control import, query and reset

bhandras:
#### Pull Request Checklist

- [x ] PR is opened against correct version branch.

### DIFF
--- a/router_client.go
+++ b/router_client.go
@@ -56,6 +56,10 @@ type RouterClient interface {
 	// currently held will be released by lnd.
 	InterceptHtlcs(ctx context.Context,
 		handler HtlcInterceptHandler) error
+
+	// ImportMissionControl imports a set of pathfinding results to lnd.
+	ImportMissionControl(ctx context.Context,
+		entries []MissionControlEntry) error
 }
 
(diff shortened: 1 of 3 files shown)


router_client.go
```diff
@@ -883,6 +886,56 @@ type MissionControlEntry struct {
...
+
+		nodeTo, err := route.NewVertexFromBytes(pair.NodeTo)
+		if err != nil {
+			return nil, err
+		}
+
+		result = append(result, MissionControlEntry{
+			NodeFrom: nodeFrom,
+			NodeTo:   nodeTo,
+			FailTime: time.Unix(pair.History.FailTime, 0),
```
carlaKC, line 926: Do we need a `!=0` check here? I think we can have entries which only have one timestamp, so we'll send up with results in 1970?

carlaKC (approved):
LGTM, just unix timestamp nit.

guggero (approved):
LGTM :tada:

router_client.go
bhandras: Yes indeed, very good catch! Fixed.

PR #83, opened 2021-12-09
lndclient: add ChanStatusFlag

yyforyongyu:
Not sure if I've got the version right. All I want to do is to include the change from `lnd`'s `4a62b7cf8c69`, which is supposed to be included in `v0.14.2` I guess.


README.md
```diff
@@ -26,7 +26,7 @@ The current compatibility matrix reads as follows:
 
 | `lndclient` git tag          | `lnd` version in `go.mod` | minimum required `lnd` version | 
 | ---------------------------- | ------------------------- | ------------------------------ |
-| [`v0.14.0-5`](https://github.com/lightninglabs/lndclient/blob/v0.14.0-5) | `v0.14.0-beta` | `v0.14.0-beta` |
+| [`v0.14.0-7`](https://github.com/lightninglabs/lndclient/blob/v0.14.0-7) | `v0.14.1-beta` | `v0.14.0-beta` |
```
carlaKC, line 29: Minimum required `lnd` version should be `lnd @ 4a62b7c`, since somebody running `v0.14.0-beta` wouldn't have this field populated in the rpc.

carlaKC (commented):
LGTM! Only question here is whether we should be opening with against a `lnd-15` branch? If this change is going to be included in `0.14.2`, then we're in the right place. But I imagine that release is going to be cherry-picked, so if it's not included then I think this belongs in 0.15?

Since we need this for the loop server, another complication is that we generally only run tagged lnd releases, so I think we should try get this into 14.2!

bhandras (approved):
LGTM :tada:

guggero:
> Only question here is whether we should be opening with against a lnd-15 branch?

Yes, I think we should do that! Or a `lnd-14-2` branch, depending on when this feature will be released in `lnd`.

README.md
yyforyongyu: Maybe we should wait for `v0.14.2-beta`?

carlaKC: We can, or make a `14-2` branch like @guggero suggested. Happy with either

yyforyongyu:
Added a new branch `lnd-14-2`. Not sure if I got it right, plz check.

README.md
guggero: With the branch we don't need to wait for the release necessary. But the README should now have its own entry for the branch with the tag `v0.14.2-0`.

guggero (approved):

README.md
yyforyongyu: cool updated!

PR #85, opened 2021-12-21
lndclient: Add chan event for ChannelEventUpdate_FULLY_RESOLVED_CHANNEL

yyforyongyu:
Depends on #83.

### DIFF
--- a/lightning_client.go
+++ b/lightning_client.go
@@ -450,6 +450,10 @@ const (
 	// InactiveChannelUpdate indicates that the channel event holds
 	// information about a channel that became inactive.
 	InactiveChannelUpdate
+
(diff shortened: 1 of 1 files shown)


guggero (approved):
Can be rebased now. If the `v0.14.2-0` tag wasn't created yet, we can also wait until this is merged and then tag it. That way we don't need `v0.14.2-1` necessarily.

yyforyongyu:
> Can be rebased now. If the `v0.14.2-0` tag wasn't created yet, we can also wait until this is merged and then tag it. That way we don't need `v0.14.2-1` necessarily.

Done!

carlaKC (approved):

PR #86, opened 2022-01-07
macaroon_service: add MacaroonService

ellemouton:
Add MacaroonService. Loop, Pool, Faraday and now Lit all need to spin up their own macaroon services and currently the logic is just duplicated in each repo. So this commit aims to consolidate all that logic into one place. 

With this commit, a caller can choose to encrypt their macaroon.db file with a password of their choosing _or_ they can specify that instead a shared key should be derived with LND which will then be used to encrypt the db.

### DIFF
--- a/macaroon_service.go
+++ b/macaroon_service.go
@@ -1,0 +1,374 @@
+package lndclient
+
+import (
+	"context"
+	"encoding/hex"
+	"errors"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"time"
+
+	"github.com/btcsuite/btcd/btcec"
+	"github.com/lightningnetwork/lnd/keychain"
+	"github.com/lightningnetwork/lnd/kvdb"
+	"github.com/lightningnetwork/lnd/lnrpc"
+	"github.com/lightningnetwork/lnd/macaroons"
+	"github.com/lightningnetwork/lnd/rpcperms"
+	"go.etcd.io/bbolt"
+	"google.golang.org/grpc"
+	"gopkg.in/macaroon-bakery.v2/bakery"
+	"gopkg.in/macaroon-bakery.v2/bakery/checkers"
+)
+
+const (
+	// defaultDBName is the default macaroon db name.
+	defaultDBName = "macaroons.db"
+
+	// defaultDBTimeout is the default timeout to be used for the
+	// macaroon db connection.
+	defaultDBTimeout = 5 * time.Second
+)
+
+var (
+	// sharedKeyNUMSBytes holds the bytes representing the compressed
+	// byte encoding of SharedKeyNUMS. It was generated via a
+	// try-and-increment approach using the phrase "Shared Secret" with
+	// SHA2-256. The code for the try-and-increment approach can be seen
+	// here: https://github.com/lightninglabs/lightning-node-connect/tree/master/mailbox/numsgen
+	sharedKeyNUMSBytes, _ = hex.DecodeString(
+		"0215b5a3e0ef58b101431e1e513dd017d1018b420bd2e89dcd71f45c031f00469e",
+	)
+
+	// SharedKeyNUMS is the public key point that can be use for when we
+	// are deriving a shared secret key with LND.
+	SharedKeyNUMS, _ = btcec.ParsePubKey(sharedKeyNUMSBytes, btcec.S256())
+
+	// SharedKeyLocator is a key locator that can be used for deriving a
+	// shared secret key with LND.
+	SharedKeyLocator = &keychain.KeyLocator{
+		Family: 21,
+		Index:  0,
+	}
+)
+
+// MacaroonService handles the creatation and unlocking of a macaroon DB file.
+// It is also a wrapper for a macaroon.Service and uses this to create a
+// default macaroon for the caller if stateless mode has not been specified.
+type MacaroonService struct {
+	cfg *MacaroonServiceConfig
+
+	db kvdb.Backend
+
+	*macaroons.Service
+}
+
+// MacaroonServiceConfig holds configuration values used by the MacaroonService.
+type MacaroonServiceConfig struct {
+	// DBPath is the path to where the macaroon db file will be stored.
+	DBPath string
+
+	// DBFile is the name of the macaroon db.
+	DBFileName string
+
+	// DBTimeout is the maximum time we wait for the bbolt database to be
+	// opened.
+	DBTimeout time.Duration
+
+	// MacaroonLocation is the value used for a macaroons' "Location" field.
+	MacaroonLocation string
+
+	// MacaroonPath is the path to where macaroons should be stored.
+	MacaroonPath string
+
+	// StatelessInit should be set to true if no default macaroons should
+	// be created and stored on disk.
+	StatelessInit bool
+
+	// Checkers are used to add extra validation checks on macaroon.
+	Checkers []macaroons.Checker
+
+	// RequiredPerms defines all method paths and the permissions required
+	// when accessing those paths.
+	RequiredPerms map[string][]bakery.Op
+
+	// Caveats is a list of caveats that will be added to the default
+	// macaroons.
+	Caveats []checkers.Caveat
+
+	// DBPassword is the password that will be used to encrypt the macaroon
+	// db. If DBPassword is not set, then LndClient, EphemeralKey and
+	// KeyLocator must be set instead.
+	DBPassword []byte
+
+	// LndClient is an LND client that can be used for any lnd queries.
+	// This only needs to be set if DBPassword is not set.
+	LndClient *LndServices
+
+	// RPCTimeout is the time after which an RPC call will be canceled if
+	// it has not yet completed.
+	RPCTimeout time.Duration
+
+	// EphemeralKey is a key that will be used to derive a shared secret
+	// with LND. This only needs to be set if DBPassword is not set.
+	EphemeralKey *btcec.PublicKey
+
+	// KeyLocator is the locator used to derive a shared secret with LND.
+	// This only needs to be set if DBPassword is not set.
+	KeyLocator *keychain.KeyLocator
+}
+
+// NewMacaroonService checks the config values passed in and creates a
+// MacaroonService object accordingly.
+func NewMacaroonService(cfg *MacaroonServiceConfig) (*MacaroonService, error) {
+	// Validate config.
+	if cfg.DBPath == "" {
+		return nil, errors.New("no macaroon db path provided")
+	}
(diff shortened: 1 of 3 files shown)


macaroon_service.go
```diff
@@ -0,0 +1,287 @@
...
+	if !cfg.StatelessInit && cfg.MacaroonPath == "" {
+		return nil, errors.New("a macaroon path must be given " +
+			"if we are not in stateless mode")
+	}
+
+	ms := MacaroonService{
+		cfg: cfg,
+	}
+
+	if cfg.DBPassword != nil {
```
guggero, line 119: Check `len()` instead?

macaroon_service.go
```diff
@@ -0,0 +1,287 @@
...
+	// EphemeralKey is a key that will be used to derive a shared secret
+	// with LND. This only needs to be set if DBPassword is not set.
+	EphemeralKey *btcec.PublicKey
+
+	// KeyLocator is the locator used to derive a shared secret with LND.
+	// This only needs to be set if DBPassword is not set.
+	KeyLocator *keychain.KeyLocator
+}
+
+func NewMacaroonService(cfg *MacaroonServiceConfig) (*MacaroonService, error) {
```
guggero, line 92: nit: Godoc

macaroon_service.go
```diff
@@ -0,0 +1,287 @@
...
+		}
+	}
+
+	return err
+}
+
+// Stop cleans up the MacaroonService.
+func (ms *MacaroonService) Stop() error {
+	var shutdownErr error
+	if err := ms.Close(); err != nil {
```
guggero, line 226: We also need to close the DB here!

macaroon_service.go
```diff
@@ -0,0 +1,287 @@
...
+
+		return nil, errors.New("must provide an LndClient, ephemeral " +
+			"key and keylocator if no DBPassword is provided " +
+			"so that a shared key can be derived with LND")
+	}
+
+	return &ms, nil
+}
+
+func (ms *MacaroonService) Start() error {
```
guggero, line 134: nit: Godoc

macaroon_service.go
```diff
@@ -0,0 +1,287 @@
...
+	db, err := kvdb.GetBoltBackend(&kvdb.BoltBackendConfig{
+		DBPath:     ms.cfg.DBPath,
+		DBFileName: ms.cfg.DBFileName,
+		DBTimeout:  ms.cfg.DBTimeout,
+	})
+	if err == bbolt.ErrTimeout {
+		return fmt.Errorf("error while trying to open %s/%s: "+
+			"timed out after %v when trying to obtain exclusive "+
+			"lock - make sure no other daemon process "+
+			"(standalone or embedded in lightning-terminal) are "+
```
guggero, line 145: nit: s/are/is ?

guggero (commented):
Very nice, this'll allow us to remove this same code from three other projects :tada: 

Looks pretty good, just a few minor comments.

macaroon_service.go
```diff
@@ -0,0 +1,287 @@
...
+	if cfg.MacaroonLocation == "" {
+		return nil, errors.New("no macaroon location provided")
+	}
+
+	if cfg.DBFileName == "" {
+		cfg.DBFileName = defaultDBName
+	}
+
+	if cfg.DBTimeout == 0 {
+		cfg.DBTimeout = defaultDBTimeout
```
carlaKC, line 107: nit: theoretically can be < 0, might want to error there

macaroon_service.go
```diff
@@ -0,0 +1,287 @@
...
+	if cfg.DBPath == "" {
+		return nil, errors.New("no macaroon db path provided")
+	}
+
+	if cfg.MacaroonLocation == "" {
+		return nil, errors.New("no macaroon location provided")
+	}
+
+	if cfg.DBFileName == "" {
+		cfg.DBFileName = defaultDBName
```
carlaKC, line 103: I'm _somewhat_ more of a fan of having a `DefaultConfig` function which sets all of these values, and then having validation just error out if the values we want aren't set, rather than mutating the config passed in. Not blocking.

macaroon_service.go
```diff
@@ -0,0 +1,287 @@
...
+	)
+	if err != nil {
+		return fmt.Errorf("unable to set up macaroon service: %v", err)
+	}
+	ms.Service = service
+
+	dbPassword := ms.cfg.DBPassword
+	if dbPassword == nil {
+		sharedKey, err := ms.cfg.LndClient.Signer.DeriveSharedKey(
+			context.Background(), ms.cfg.EphemeralKey,
```
carlaKC, line 166: do we want timeout on this ctx?

carlaKC (commented):
Looks good! 

Would just bump one of the dependent services to use this before merge to test it out. There's inevitably something small that needs fixing, so nice to see before merge.

macaroon_service.go
ellemouton: do you mean: generate default config, then checking each of the passed in config values to see if the default config values should be overwritten and then returning the modified default config struct? feels like that is more bloated cause would then need to check every single config item to see if it has changed from the default. wdyt?

macaroon_service.go
ellemouton: added. thanks!

macaroon_service.go
guggero: I think a default config works well when using CLI flags, as the user is explicitly overwriting values. But since this is a library function, using a default struct does make it more bloated. So I think this is fine the way it is.

macaroon_service.go
```diff
@@ -0,0 +1,315 @@
...
+	if !cfg.StatelessInit && cfg.MacaroonPath == "" {
+		return nil, errors.New("a macaroon path must be given if we " +
+			"are not in stateless mode")
+	}
+
+	ms := MacaroonService{
+		cfg: cfg,
+	}
+
+	if len(cfg.DBPassword) == 0 {
```
guggero, line 135: Should be `!= 0`. Though looking at the Pool PR, do we need a way to migrate the DB to a non-empty password?
We could add a special case here. If we have an empty password but all the other variables set (lnd client, ephemeral key and key locator), we first try to unlock with the shared key (in case migration was successful already). If that fails, try to unlock with an empty password. If that succeeds, change the password to the shared key.

Though that would make this a breaking change for Loop/Pool/Faraday, as users couldn't go back to an older version after migrating the DB. I think that is fine, since they could just delete the macaroon DB if they really needed to go back to a previous version.

guggero (commented):
One last comment about migrating away from an empty password. That might be worthy of a unit test too.

macaroon_service.go
ellemouton: > Should be != 0

right. i just realised that the reason i initially had it as `if cfg.DBPassword != nil` was cause currently pool & loop set zero len (but non-nil) db passwords here. 

>  If we have an empty password but all the other variables set (lnd client, ephemeral key and key locator), we first try to unlock with the shared key (in case migration was successful already). If that fails, try to unlock with an empty password. If that succeeds, change the password to the shared key.

cool, sounds good!

ellemouton:
Thanks for the reviews @carlaKC  & @guggero 🚀 

I have updated it as @guggero has suggested to allow for migration.  However, now it does not allow the client to have an empty passphrase at all. Is that wanted? or do we still want to accommodate that case? if so, i can easily update it to include that.

I will re-request review once i have tested the updates properly

guggero:
> However, now it does not allow the client to have an empty passphrase at all. Is that wanted?

Yes, I think a passphrase should be mandatory so the DB is encrypted. Either it's supplied as a string or derived through the shared secret.

ellemouton:
Ok i think this is ready for another round now 👍 

see [this](https://github.com/lightninglabs/pool/pull/336) Pool PR to see how this is used there

carlaKC (approved):
LGTM, tested with loop PR and all in order 👌

macaroon_service.go
```diff
@@ -0,0 +1,364 @@
...
+var (
+	// MacaroonPubKey is the public key point that can be use for when we
+	// are deriving a shared secret key with LND. It was generated via a
+	// try-and-increment approach using the phrase "Shared Secret" with
+	// SHA2-256. The code for the try-and-increment approach can be seen
+	// here: https://github.com/lightninglabs/lightning-node-connect/tree/master/mailbox/numsgen
+	nBytes, _ = hex.DecodeString(
+		"0215b5a3e0ef58b101431e1e513dd017d1018b420bd2e89dcd71f45c031f00469e",
+	)
+	MacaroonPubKey, _ = btcec.ParsePubKey(nBytes, btcec.S256())
```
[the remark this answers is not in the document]
guggero: How about `SharedKeyNUMS` (=nothing up my sleeve, nobody knows the private key for this point)?

macaroon_service.go
```diff
@@ -0,0 +1,364 @@
...
+	defaultDBTimeout = 5 * time.Second
+)
+
+var (
+	// MacaroonPubKey is the public key point that can be use for when we
+	// are deriving a shared secret key with LND. It was generated via a
+	// try-and-increment approach using the phrase "Shared Secret" with
+	// SHA2-256. The code for the try-and-increment approach can be seen
+	// here: https://github.com/lightninglabs/lightning-node-connect/tree/master/mailbox/numsgen
+	nBytes, _ = hex.DecodeString(
```
guggero, line 39: nit: `sharedKeyNUMSBytes` and split godoc comment? Perhaps the `..Bytes` variable can say how it was generated and the actual pubkey for what it could be used?

macaroon_service.go
```diff
@@ -0,0 +1,364 @@
...
+		"0215b5a3e0ef58b101431e1e513dd017d1018b420bd2e89dcd71f45c031f00469e",
+	)
+	MacaroonPubKey, _ = btcec.ParsePubKey(nBytes, btcec.S256())
+)
```
guggero, line 43: Let's also add a `SharedKeyLocator` variable here that can be used by all the daemons:

```
	SharedKeyLocator = &keychain.KeyLocator{
		Family: 21,
		Index:  0,
	}
```

guggero (approved):
Very nice work on this, turned out great!

Two last nits, then we can get this merged :tada:

ellemouton:
Shweeeet, thanks @guggero 🚀 updated 👍 
And is this opened to the correct branch? should it be opened to 0.14.2 rather?

guggero:
This PR doesn't use anything that was only introduced in 0.14.2, so 0.14.0 is correct.
I'll create and push the tag `v0.14.0-8` after merging this PR.

PR #87, opened 2022-02-04
Add StopDaemon rpc.

sebdeveloper6952:
#### Pull Request Checklist

- [ x ] PR is opened against correct version branch.
- [ x ] Version compatibility matrix in the README and minimal required version
- [ x ] Update `macaroon_recipes.go` if your PR adds a new method that is called

### DIFF
--- a/lightning_client.go
+++ b/lightning_client.go
@@ -198,6 +198,10 @@ type LightningClient interface {
 	RegisterRPCMiddleware(ctx context.Context, middlewareName,
 		customCaveatName string, readOnly bool, timeout time.Duration,
 		intercept InterceptFunction) (chan error, error)
+
+	// StopDaemon will send a shutdown request to the interrupt handler,
+	// triggering a graceful shutdown of the daemon.
+	StopDaemon(ctx context.Context) error
 }
 
 // Info contains info about the connected lnd node.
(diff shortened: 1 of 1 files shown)


lightning_client.go
```diff
@@ -3689,3 +3693,18 @@ func (s *lightningClient) RegisterRPCMiddleware(ctx context.Context,
 
 	return errChan, nil
 }
+
+func (s *lightningClient) StopDaemon(ctx context.Context) error {
```
guggero, line 3697: nit: add same Godoc comment here.

lightning_client.go
```diff
@@ -3689,3 +3693,18 @@ func (s *lightningClient) RegisterRPCMiddleware(ctx context.Context,
...
+	rpcCtx, cancel := context.WithTimeout(ctx, s.timeout)
+	defer cancel()
+
+	rpcCtx = s.adminMac.WithMacaroonAuth(rpcCtx)
+	rpcReq := &lnrpc.StopRequest{}
+
+	_, err := s.client.StopDaemon(rpcCtx, rpcReq)
+	if err != nil {
```
guggero, line 3705: nit: return err directly.

guggero (changes requested):
Thank you for the PR!
Since this RPC existed in `lnd` for a long time, I think we should open this against the `master` branch (currently tracking 0.13.x) and then rebase all other branches.

sebdeveloper6952:
Thank you for the suggestions @guggero!

PR #88, opened 2022-02-17
lndclient: add closing txid for waiting close channel response

yyforyongyu:
Bump lnd's version to v0.14.2-beta to include the closing txid from waiting close channel response.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ The current compatibility matrix reads as follows:
 
 | `lndclient` git tag          | `lnd` version in `go.mod` | minimum required `lnd` version | 
 | ---------------------------- | ------------------------- | ------------------------------ |
-| [`v0.14.2-0`](https://github.com/lightninglabs/lndclient/blob/v0.14.2-0) | `lnd @ 4a62b7cf8c69` | `v0.14.0-beta` |
+| [`v0.14.2-1`](https://github.com/lightninglabs/lndclient/blob/v0.14.2-1) | `v0.14.2-beta` | `v0.14.2-beta` |
 | [`v0.13.0-10`](https://github.com/lightninglabs/lndclient/blob/v0.13.0-10) | `lnd @ d9f0f07142ea` | `v0.13.0-beta` |
 | `master` / [`v0.12.0-13`](https://github.com/lightninglabs/lndclient/blob/v0.12.0-13) | `v0.12.0-beta` | `v0.12.0-beta` |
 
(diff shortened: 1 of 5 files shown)


carlaKC (commented):
Should this be on the 14.2 branch? Looks like the relevant [commit](https://github.com/lightningnetwork/lnd/commit/99378b22c3a3b9b7f565f07a2241042c2d6477ec) wasn't included in the release? So I think this should go on a .15 branch, since that's the next release that'll have the field?

README.md
```diff
@@ -26,7 +26,7 @@ The current compatibility matrix reads as follows:
 
 | `lndclient` git tag          | `lnd` version in `go.mod` | minimum required `lnd` version | 
 | ---------------------------- | ------------------------- | ------------------------------ |
-| [`v0.14.2-0`](https://github.com/lightninglabs/lndclient/blob/v0.14.2-0) | `lnd @ 4a62b7cf8c69` | `v0.14.0-beta` |
+| [`v0.14.2-1`](https://github.com/lightninglabs/lndclient/blob/v0.14.2-1) | `v0.14.2-beta` | `lnd @ 99378b22c3a3` |
```
guggero, line 29: nit: can remove/replace the `lnd@...` string and just use `v0.14.2-beta`?

lightning_client.go
```diff
@@ -1845,12 +1848,18 @@ func (s *lightningClient) PendingChannels(ctx context.Context) (*PendingChannels
 			return nil, err
 		}
 
+		hash, err := chainhash.NewHashFromStr(waiting.ClosingTxid)
```
guggero, line 1851: Are we sure this is always set? If this was only added in 0.14.2 then we'd need to bump the `minimalCompatibleVersion` to make sure we never run into an error here.

yyforyongyu: Fixed.

README.md
yyforyongyu: Done.

yyforyongyu:
> Should this be on the 14.2 branch? Looks like the relevant [commit](https://github.com/lightningnetwork/lnd/commit/99378b22c3a3b9b7f565f07a2241042c2d6477ec) wasn't included in the release? So I think this should go on a .15 branch, since that's the next release that'll have the field?

It's there actually, but weirdly in a different [commit](https://github.com/lightningnetwork/lnd/commit/981464846a16b92eb02ce19dd09cd125b364814b). Plus the build won't pass if it's not there.

guggero (approved):

carlaKC (approved):

PR #99, opened 2022-05-07
macaroons: update to new macaroons pkg version

Roasbeef:
In this commit we update to the new version of the macaroons package
that doesn't require the caller to use the default implementation for
the root key store. Instead, the caller can pass in an implementation,
and based on that we'll change our behavior.

#### Pull Request Checklist

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -11,7 +11,6 @@ require (
 	github.com/lightningnetwork/lnd v0.15.0-beta.rc6.0.20220707020049-dc35f78ebce6
 	github.com/lightningnetwork/lnd/kvdb v1.3.1
 	github.com/stretchr/testify v1.7.1
-	go.etcd.io/bbolt v1.3.6
 	google.golang.org/grpc v1.38.0
 	gopkg.in/macaroon-bakery.v2 v2.0.1
 	gopkg.in/macaroon.v2 v2.1.0
@@ -109,6 +108,7 @@ require (
 	github.com/ulikunitz/xz v0.5.10 // indirect
 	github.com/xi2/xz v0.0.0-20171230120015-48954b6210f8 // indirect
 	github.com/xiang90/probing v0.0.0-20190116061207-43a291ad63a2 // indirect
+	go.etcd.io/bbolt v1.3.6 // indirect
 	go.etcd.io/etcd/api/v3 v3.5.0 // indirect
 	go.etcd.io/etcd/client/pkg/v3 v3.5.0 // indirect
 	go.etcd.io/etcd/client/v2 v2.305.0 // indirect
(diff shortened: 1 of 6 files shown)


go.mod
```diff
@@ -1,19 +1,20 @@
...
-	go.etcd.io/bbolt v1.3.6
+	github.com/stretchr/testify v1.7.1
 	google.golang.org/grpc v1.38.0
 	gopkg.in/macaroon-bakery.v2 v2.0.1
 	gopkg.in/macaroon.v2 v2.1.0
 )
 
 go 1.16
+
+replace github.com/lightningnetwork/lnd => github.com/roasbeef/lnd v0.2.1-alpha.0.20220507000000-660f73377f6f
```
Roasbeef, line 20: Will be removed once the lnd PR is merged.

macaroon_service.go
```diff
@@ -170,40 +150,27 @@ func NewMacaroonService(cfg *MacaroonServiceConfig) (*MacaroonService, error) {
...
-// Start starts the macaroon validation service, creates or unlocks the macaroon
-// database and, if we are not in stateless mode, creates the default macaroon
-// if it doesn't exist yet.
+// Start starts the macaroon validation service, creates or unlocks the
+// macaroon database and, if we are not in stateless mode, creates the default
+// macaroon if it doesn't exist yet.
 func (ms *MacaroonService) Start() error {
-	// Open the database.
-	var err error
-	ms.db, err = kvdb.GetBoltBackend(&kvdb.BoltBackendConfig{
```
guggero, line 179: Can we extract this into a convenience function please? Otherwise we need to duplicate this code in all projects (loop/pool/faraday/litd).

macaroon_service_test.go
```diff
@@ -24,14 +25,21 @@ func TestMacaroonServiceMigration(t *testing.T) {
 	require.NoError(t, err)
 	defer os.RemoveAll(tempDirPath)
 
+	db, err := kvdb.GetBoltBackend(&kvdb.BoltBackendConfig{
+		DBPath:     tempDirPath,
+		DBFileName: "macaroons.db",
+		DBTimeout:  defaultDBTimeout,
+	})
+	require.NoError(t, err)
+	rks, err := macaroons.NewRootKeyStorage(db)
```
guggero, line 34: The convenience function mentioned above would be handy here.

guggero (approved):
LGTM, just one minor request.

Roasbeef:
SGTM re extracting them, I have a new daemon using this fully working, so can confirm all the changes allow one to import this and have it handle the initial macaroon creation and rook key creation as well.

macaroon_service_test.go
Roasbeef: Added!

Roasbeef:
Updated to target the 0.15 branch since that has everything the 0.16 one did. I also updated the permissions file, to then add `ImportPublicKey` as well.

PR #101, opened 2022-05-16
client: add a more generic `EstimateFee` and remove `EstimateFeeP2WSH`

bhandras:
#### Pull Request Checklist

### DIFF
--- a/lightning_client.go
+++ b/lightning_client.go
@@ -36,6 +36,11 @@ type LightningClient interface {
 
 	GetInfo(ctx context.Context) (*Info, error)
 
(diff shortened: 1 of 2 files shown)


bhandras:
Opened againsg `lnd-15-0` but we could technically open it to `master` too. The context is to make `EstimateFee` work with taproot addresses too.

guggero (approved):

guggero:
I think 0.15 is okay, since this is a breaking change (meaning it requires code changes for users of the library but will remain backward compatible of course).

PR #105, opened 2022-06-09
multi: add PSBT methods to wallet kit

guggero:
#### Pull Request Checklist


testdata/permissions.json
```diff
@@ -868,6 +868,14 @@
                 }
             ]
         },
+        "/walletrpc.WalletKit/SignPsbt": {
```
guggero, line 871: The `FundPsbt` and `FinalizePsbt` methods were already here.

walletkit_client.go
```diff
@@ -407,3 +458,135 @@ func (m *walletKitClient) ListAccounts(ctx context.Context, name string,
...
+	}
+
+	finalizedPacket, err := psbt.NewFromRawBytes(
+		bytes.NewReader(resp.SignedPsbt), false,
+	)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	finalTx := wire.NewMsgTx(2)
```
positiveblue, line 587: out of scope for this PR but we may want to create some iota for the tx version?

walletkit_client.go
```diff
@@ -407,3 +460,135 @@ func (m *walletKitClient) ListAccounts(ctx context.Context, name string,
...
+	defer cancel()
+
+	resp, err := m.client.FundPsbt(
+		m.walletKitMac.WithMacaroonAuth(rpcCtx), req,
+	)
+	if err != nil {
+		return nil, 0, nil, err
+	}
+
+	packet, err := psbt.NewFromRawBytes(
```
bhandras, line 494: I suppose this is the kind of error that must never happen since otherwise the selected inputs will be locked forever?

bhandras (approved):
LGTM 👌

walletkit_client.go
guggero: Hmm, it looks like we're only using it this way throughout the lnd codebase. But you're right, we should probably do it at some point :)

walletkit_client.go
guggero: Yeah... Though the inputs wouldn't be locked forever, only 10 minutes.

PR #112, opened 2022-07-08
client: Add Custom Messaging APIs

carlaKC:
This PR adds Send/Subscribe custom message wrappers to lightning_client.

#### Pull Request Checklist

       ->  Master is on lnd 0.14.3, custom messages were merged in 0.14.0
- [ ] Version compatibility matrix in the README and minimal required version in `lnd_services.go` are updated.
       -> Change only needs 0.14, so no update required. 
- [ ] Update `macaroon_recipes.go` if your PR adds a new method that is called differently than the RPC method it invokes.
       -> [SendCustomMessage](https://api.lightning.community/#sendcustommessage) and [SubscribeCustomMessage](https://api.lightning.community/#subscribecustommessages) implemented have the same names and lnd's API.


lightning_client.go
```diff
@@ -3757,3 +3832,84 @@ func (s *lightningClient) SendCustomMessage(ctx context.Context,
...
+	)
+
+	s.wg.Add(1)
+	go func() {
+		defer s.wg.Done()
+
+		for {
+			// Before we try to receive, do a quick sanity check
+			// that our context wasn't canceled by the caller. This
+			// allows us to exist faster than waiting for Recv() to
```
bhandras, line 3873: I like to exist faster :)

guggero (approved):

lightning_client.go
```diff
@@ -1091,6 +1097,75 @@ type CustomMessage struct {
...
+	// If we have a peer filter set, only return messages if their peer is
+	// in our map.
+	if c.PeerFilter != nil {
+		_, peerIncluded := c.PeerFilter[msg.Peer]
+		include = include && peerIncluded
+	}
+
+	// If we have a type filter set, only return messages if their type is
+	// in our filter map.
+	if c.TypeFilter != nil {
```
bhandras, line 1126: nit: could check `include &&` here.

lightning_client.go
```diff
@@ -1091,6 +1097,75 @@ type CustomMessage struct {
 	Data []byte
 }
 
+// CustomMessageFilter provides a filter for subscribing to custom messages.
+type CustomMessageFilter struct {
```
bhandras, line 1101: This is a nice abstraction, but not sure if we should put this in `lndclient` as it's more application related imo. Some users may create positive filters some may create negative filters, etc.

bhandras (commented):
Cool PR, looks good, I'm just not sure about the `CustoMessageFilter`.

lightning_client.go
carlaKC: Good point, I was on the fence about it too. Looking at this again, if we wanted a filter it could be a `FilterMessage(CustomMessage)` closure that you can pass in, rather than fixing the structure in here. 

Less code/more good, so will remove!

bhandras: Yeah a closure could work just fine. Avoids waking up whoever is waiting on the channel and gives more opportunity for customization.

lightning_client.go
```diff
@@ -3757,3 +3762,60 @@ func (s *lightningClient) SendCustomMessage(ctx context.Context,
 	_, err := s.client.SendCustomMessage(rpcCtx, rpcReq)
 	return err
 }
+
+// SubscribeCustomMessages subscribes to a stream of custom messages, optionally
+// filtering by peer and message type.
+func (s *lightningClient) SubscribeCustomMessages(ctx context.Context) (
+	chan<- CustomMessage, chan<- error, error) {
+
+	rpcCtx, cancel := context.WithTimeout(ctx, s.timeout)
```
bhandras, line 3771: Not sure about this timeout context, will this also abort the stream?

lightning_client.go
```diff
@@ -3757,3 +3762,60 @@ func (s *lightningClient) SendCustomMessage(ctx context.Context,
...
+	)
+
+	s.wg.Add(1)
+	go func() {
+		defer s.wg.Done()
+
+		for {
+			msg, err := client.Recv()
+			if err != nil {
+				errChan <- fmt.Errorf("receive failed: %w", err)
```
bhandras, line 3796: Instead of directly sending to the `errChan` we could use similar pattern as when we're writing to the `msgChan` to avoid a potential deadlock and a lingering goroutine.

```
sendErr := func(err error) {
		select {
		case errChan <- err:
		case <-ctx.Done():
		}
	}
```

lightning_client.go
carlaKC: Yeah indeed, nice catch!

lightning_client.go
carlaKC: I don't think we're at risk of deadlock with `errChan` buffered by 1? Since this loop will only ever send 1 or 0 errors into it then exit?

carlaKC:
Push diff [here](https://github.com/lightninglabs/lndclient/compare/e7e44d68572a2d3e7060edeb91268d773ea82d1e..988c050334961bb3e18e0127da31b1681ad4a4ae), removed incorrect timeout on subscritpion ctx and close channels on subscription close (to signal to callers that we're done).

I've successfully used `SendCustomMessage` in my own code, will do the same with `SubscribeCustomMessage` and drop a comment here before re-requesting so that we know there are no other silly bugs lingering like ctx timeout :')

lightning_client.go
bhandras: Yes indeed, missed that it's buffered.

lightning_client.go
```diff
@@ -3757,3 +3762,62 @@ func (s *lightningClient) SendCustomMessage(ctx context.Context,
...
+		// channel does not block our exit.
+		errChan = make(chan error, 1)
+		msgChan = make(chan CustomMessage)
+	)
+
+	s.wg.Add(1)
+	go func() {
+		defer func() {
+			s.wg.Done()
+			close(errChan)
```
bhandras, line 3791: nit: maybe we could change the order, close channels first then waitgroup.

bhandras (approved):
LGTM :100:

lightning_client.go
carlaKC: [Changed](https://github.com/lightninglabs/lndclient/compare/ebeabd076237480658be58ec122132809eb9d53a..a792c60eb81aa94d804ace90bff3e95eb9b5ceb9) 👍

PR #113, opened 2022-07-14
lndclient: expose new option to recv the block w/ a conf ntfn

Roasbeef:
In order to not break all the other callers, we opt to more or less copy
over the functional options added on the lnd side of things. We would be
able to just use the existing opts, but the includeBlock field is
private (as it should be imo).

### DIFF
--- a/chainnotifier_client.go
+++ b/chainnotifier_client.go
@@ -13,13 +13,39 @@ import (
 	"google.golang.org/grpc"
 )
 
+// notifierOptions is a set of functional options that allow callers to further
+// modify the type of chain even notifications they receive.
+type notifierOptions struct {
+	// includeBlock if true, then the dispatched confirmation notification
+	// will include the block that mined the transaction.
+	includeBlock bool
+}
+
+// defaultNotifierOptions returns the set of default options for the notifier.
+func defaultNotifierOptions() *notifierOptions {
+	return &notifierOptions{}
+}
+
+// NotifierOption is a functional option that allows a caller to modify the
+// events received from the notifier.
+type NotifierOption func(*notifierOptions)
+
+// WithIncludeBlock is an optional argument that allows the calelr to specify
+// that the block that mined a transaction should be included in the response.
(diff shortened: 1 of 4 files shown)


Roasbeef:
Switched to target the lnd-16-0 branch.

chainnotifier_client.go
```diff
@@ -13,13 +13,40 @@ import (
 	"google.golang.org/grpc"
 )
 
+// notifierOptions is a set of functional options that allow callers to further
+// modify the type of chain even notifications they receive.
+type notifierOptions struct {
+	// includeBlock if true, then the dispatched confirmation notification
+	// will include the block that mined the transaction.  includeBlock
+	// bool
```
carlaKC, line 21: extra `includeBlock // bool` here

chainnotifier_client.go
```diff
@@ -170,18 +204,27 @@ func (s *chainNotifierClient) RegisterConfirmationsNtfn(ctx context.Context,
 					errChan <- err
 					return
 				}
+
+				block, err := decodeBlock(c.Conf.RawBlock)
```
carlaKC, line 208: If `includeBlock=false`, wouldn't `c.Conf.RawBlock` be empty / fail deserialization in `decodeBlock`?

guggero: Yes, I think that should be guarded by another `if` or return early in `decodeBlock`.

guggero (approved):
Can be rebased onto `lnd-16-0` (which I rebased on top of the `lnd-15-0` and `lnd-15-1` branches).

LGTM pending @carlaKC's comments.

Roasbeef:
Ok it should be on the proper branch now. Still needs the `lnd` PR to land before this can.

guggero:
Can be rebased and updated now that https://github.com/lightningnetwork/lnd/pull/6730 is merged.

Roasbeef:
PR is now full rebased!

PR #116, opened 2022-08-03
Add support for AMP and zero amount invoices

thomasbarrett:
#### Pull Request Checklist


thomasbarrett:
Hey! I am trying to use LndClient to do real-time payment streaming with a static AMP invoice. I ran into two issues using lndclient v0.16.0.

1. Support for AMP payments does not seem to be built out yet in the AddInvoice method. Notably, the IsAmp field is not properly set in the `lnrpc.Invoice` object.

2. Zero value invoices are not properly supported in the SendPayment method. Currently, a Value is only forwarded if the Invoice is not set. Ideally, we would forward the Value independently of whether the Invoice is set to enable zero-value invoices.

guggero:
Hey. I agree with the `IsAmp` flag, that needs to be added to the `AddInvoice` call. Though I'm not sure about the `Amt` field. It might be confusing to always add the value, especially if it gets overwritten if an invoice is also passed in.
How about we add a new field to the `SendPaymentRequest` called `AmpAmount` that is only set if the invoice has a zero amount. Would that work for your use case?

PR #121, opened 2022-08-18
fix: revert changes that are not lnd v.0.15.0-beta compatible

positiveblue:
Reverts #99

### DIFF
--- a/testdata/permissions.json
+++ b/testdata/permissions.json
@@ -64,14 +64,6 @@
                 }
             ]
         },
-        "/devrpc.Dev/ImportGraph": {
-            "permissions": [
-                {
-                    "entity": "offchain",
-                    "action": "write"
-                }
-            ]
-        },
         "/invoicesrpc.Invoices/AddHoldInvoice": {
             "permissions": [
                 {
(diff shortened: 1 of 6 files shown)


testdata/permissions.json
```diff
@@ -64,14 +64,6 @@
                 }
             ]
         },
-        "/devrpc.Dev/ImportGraph": {
```
guggero, line 67: This commit can stay, this just adds things that should've already been there.

walletkit_client.go
```diff
@@ -130,10 +130,6 @@ type WalletKitClient interface {
 	// this method.
 	FinalizePsbt(ctx context.Context, packet *psbt.Packet,
 		account string) (*psbt.Packet, *wire.MsgTx, error)
-
-	// ImportPublicKey imports a public key as watch-only into the wallet.
-	ImportPublicKey(ctx context.Context, pubkey *btcec.PublicKey,
```
guggero, line 135: This commit can be kept as well, all of this is in `0.15.0-beta`.

guggero (changes requested):
We should only revert the macaroon and lnd version related commits (3 of them), the rest are okay to remain on master.

guggero:
Since I broke this by merging the 0.15 branch, I fixed it now by cherry-picking the correct commits from this PR. Thanks a lot for getting things started!

PR #124, opened 2022-09-06
lndclient: add WithFundingShim opt to openchannel

GeorgeTsagk:
#### Pull Request Checklist

#### Description

Adds the ability to attach funding shims to open channel requests. It is introduced as an optional argument.


bhandras (approved):
LGTM :+1:

PR #126, opened 2022-09-12
lndclient: add FundingStateStep call

GeorgeTsagk:
#### Pull Request Checklist

#### Description

Adds a new wrapper for the LND RPC call [`FundingStateStep`](https://api.lightning.community/#fundingstatestep) as a method to `LightningClient`.

### DIFF
--- a/lightning_client.go
+++ b/lightning_client.go
@@ -186,6 +186,12 @@ type LightningClient interface {
 	ChannelAcceptor(ctx context.Context, timeout time.Duration,
 		accept AcceptorFunction) (chan error, error)
 
+	// FundingStateStep is a funding related call that allows the execution
+	// of some preparatory steps for a funding workflow or manual
+	// progression of a funding workflow.
+	FundingStateStep(ctx context.Context, req *lnrpc.FundingTransitionMsg) (
+		*lnrpc.FundingStateStepResp, error)
+
 	// QueryRoutes can query LND to return a route (with fees) between two
 	// vertices.
 	QueryRoutes(ctx context.Context, req QueryRoutesRequest) (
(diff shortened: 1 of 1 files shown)


positiveblue (approved):
LGTM :+1:

lightning_client.go
```diff
@@ -3619,6 +3625,20 @@ func (s *lightningClient) ChannelAcceptor(ctx context.Context,
 	return errChan, nil
 }
 
+// FundingStateStep is a funding related call that allows the execution
+// of some preparatory steps for a funding workflow or manual
+// progression of a funding workflow.
+func (s *lightningClient) FundingStateStep(ctx context.Context,
```
bhandras, line 3631: I'm wondering if it's better to keep this call as a wrapper around the raw rpc client call or create some higher level abstraction that fits golang better? Maybe @guggero can also chime in here since funding is his expertise.

lightning_client.go
```diff
@@ -3619,6 +3625,20 @@ func (s *lightningClient) ChannelAcceptor(ctx context.Context,
...
+	req *lnrpc.FundingTransitionMsg) (*lnrpc.FundingStateStepResp, error) {
+
+	resp, err := s.client.FundingStateStep(ctx, req)
```
bhandras, line 3634: In order to make this call succeed you also need to wrap the context, first with a timeout context then also add permissions:

```
rpcCtx, cancel := context.WithTimeout(ctx, s.timeout)
defer cancel()
rpcCtx = s.adminMac.WithMacaroonAuth(rpcCtx)
```

bhandras (approved):
Current change LGTM, awaiting @guggero 's opinion.

lightning_client.go
guggero: I think it's okay to have the light wrapper as it's currently implemented in this PR. Otherwise we'd need an enum or multiple structs to abstract the different messages that can be sent as a funding state step. And I don't really see the benefit in that.
We probably should just add all existing RPCs to `lndclient` so we can deprecate the use of the basic client.

guggero (changes requested):
Since this call has existed for a long time, this PR should be made to the `master` branch instead.
Code looks good to me though :tada:

guggero (approved):

PR #128, opened 2022-09-14
lndclient: add openchannel async call

GeorgeTsagk:
#### Pull Request Checklist

#### Description

Adds a new method for asynchronous `OpenChannel`. It returns the grpc stream for the `lndclient` user to listen to updates.

### DIFF
--- a/lightning_client.go
+++ b/lightning_client.go
@@ -153,6 +153,14 @@ type LightningClient interface {
 		opts ...OpenChannelOption) (
 		*wire.OutPoint, error)
 
+	// OpenChannelStream opens a channel to the specified peer and with the
+	// specified arguments and options. This function returns a stream of
+	// updates.
+	OpenChannelStream(ctx context.Context, peer route.Vertex,
+		localSat, pushSat btcutil.Amount, private bool,
+		opts ...OpenChannelOption) (<-chan *OpenStatusUpdate,
+		<-chan error, error)
+
 	// CloseChannel closes the channel provided.
 	CloseChannel(ctx context.Context, channel *wire.OutPoint,
 		force bool, confTarget int32, deliveryAddr btcutil.Address) (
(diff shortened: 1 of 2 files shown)


lightning_client.go
```diff
@@ -2733,6 +2741,37 @@ func (s *lightningClient) OpenChannel(ctx context.Context, peer route.Vertex,
...
 }
 
+// OpenChannelStream opens a channel to the specified peer and with the
+// specified arguments and options. This function returns a stream of
+// updates.
+func (s *lightningClient) OpenChannelStream(ctx context.Context, peer route.Vertex,
+	localSat, pushSat btcutil.Amount, private bool,
+	opts ...OpenChannelOption) (lnrpc.Lightning_OpenChannelClient, error) {
+
+	rpcCtx, cancel := context.WithTimeout(ctx, s.timeout)
```
guggero, line 2751: We can't use a timeout here, since the stream might be open for a very long time (if the subscriber wants to keep it open until the confirmation is received).

lightning_client.go
```diff
@@ -153,6 +153,14 @@ type LightningClient interface {
 		opts ...OpenChannelOption) (
 		*wire.OutPoint, error)
 
+	// OpenChannelStream opens a channel to the specified peer and with the
+	// specified arguments and options. This function returns a stream of
+	// updates.
+	OpenChannelStream(ctx context.Context, peer route.Vertex,
```
guggero, line 159: I think this also requires an entry in the `renames` map of the `macaroon_recipes.go` file. Though I wonder why that test didn't fail here...

lightning_client.go
guggero: This is the main difference to the `*Sync` call btw., that only waits for a `lnrpc.OpenStatusUpdate_ChanPending` message while here we might want to wait for `OpenStatusUpdate_ChanOpen` or `OpenStatusUpdate_PsbtFund` messages.

lightning_client.go
```diff
@@ -2733,6 +2741,37 @@ func (s *lightningClient) OpenChannel(ctx context.Context, peer route.Vertex,
...
+		Private:            private,
+	}
+
+	for _, opt := range opts {
+		opt(rpcRequest)
+	}
+
+	rpcCtx = s.adminMac.WithMacaroonAuth(rpcCtx)
+
+	stream, err := s.client.OpenChannel(rpcCtx, rpcRequest)
```
guggero, line 2767: I guess the "main purpose" of `lndclient` is to abstract away some of the gRPC stream handling logic. So it would make sense to return native Golang channels here instead of the raw stream. See `SubscribeChannelEvents` for an example.

lightning_client.go
```diff
@@ -520,6 +528,25 @@ const (
 	FullyResolvedChannelUpdate
 )
 
+type OpenStatusUpdate struct {
```
GeorgeTsagk, line 531: @guggero I did not include an `enum` on the `OpenStatusUpdate` struct as a non-nil field indicates the type of the update.

guggero: SGTM :+1: 
Two nits here: The struct itself needs a Godoc comment. And the comments for the fields aren't valid Godoc comments as they need to start with the field name.

guggero (approved):
LGTM, just two small nits regarding Godoc comments.

lightning_client.go
```diff
@@ -2733,6 +2764,78 @@ func (s *lightningClient) OpenChannel(ctx context.Context, peer route.Vertex,
...
+		LocalFundingAmount: int64(localSat),
+		PushSat:            int64(pushSat),
+		Private:            private,
+	}
+
+	for _, opt := range opts {
+		opt(rpcRequest)
+	}
+
+	updateStream, err := s.client.OpenChannel(ctx, rpcRequest)
```
bhandras, line 2807: Shouldn't we use the decorated rpc context here?

PR #130, opened 2022-10-26
lightning-client: return correct context error

ellemouton:
This commit fixes a bug where the in-correct context's error message is returned resulting in a nil error being returned when the context is canceled. On a middleware registration error, the first thing we will pick up here is that the context is cancelled, however, this does not provide the user with much feedback about what went wrong and so an additional log statement is also added on the stream's Recv func as this will provide the user with more info.


guggero (approved):
Nice catch, LGTM :tada:

PR #131, opened 2022-11-01
mod: bump btcd module to v0.23.3

positiveblue:
#### Pull Request Checklist


guggero:
Not sure we use a `v0.15.0-X` version anywhere in our projects? So maybe this needs to be against a branch instead of `master`?

PR #132, opened 2022-11-03
mod: bump btcd to v0.23.3 and lnd to v0.15.4-beta

positiveblue:
#### Pull Request Checklist


go.mod
```diff
@@ -1,14 +1,14 @@
...
-	github.com/btcsuite/btcd v0.23.1
+	github.com/btcsuite/btcd v0.23.3
 	github.com/btcsuite/btcd/btcec/v2 v2.2.1
 	github.com/btcsuite/btcd/btcutil v1.1.2
 	github.com/btcsuite/btcd/btcutil/psbt v1.1.5
 	github.com/btcsuite/btcd/chaincfg/chainhash v1.0.1
 	github.com/btcsuite/btclog v0.0.0-20170628155309-84c8d2346e9f
 	github.com/btcsuite/btcwallet/wtxmgr v1.5.0
-	github.com/lightningnetwork/lnd v0.15.1-beta
+	github.com/lightningnetwork/lnd v0.15.4-beta
```
bhandras, line 11: I think it'd be better to create an `lnd-15-4` branch and open this PR against that? Otherwise we end up with an `lnd-15-1` branch that requires 0.15.4.

guggero: Since we're on a 0.15.x version in all our projects and 0.15.4 is the global default minimum, I think it actually makes sense to merge the `0-15-1` branch to master and update master to 0.15.4. I'll do that and then you can set the base branch to `master`, @positiveblue.

PR #133, opened 2022-11-21
client: add more detail to the pending channels

bhandras:
This PR adds more detail to the pending channels. Local and remote balances and force close maturity, anchor state, limbo and recovered balances.

#### Pull Request Checklist

### DIFF
--- a/lightning_client.go
+++ b/lightning_client.go
@@ -683,6 +683,24 @@ const (
 	InitiatorBoth
 )
 
+// ForceCloseAnchorState indicates the resolution state for the anchor.
+// There are three resolution states for the anchor: limbo, lost and recovered.
+type ForceCloseAnchorState int32
+
+const (
+	// ForceCloseAnchorStateLimbo is set if the recovered_balance is zero
+	// and limbo_balance is non-zero.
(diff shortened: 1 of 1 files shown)


lightning_client.go
```diff
@@ -683,6 +683,24 @@ const (
...
 )
 
+// ForceCloseAnchorState indicates the resolution state for the anchor.
+// There are three resolution states for the anchor: limbo, lost and recovered.
+type ForceCloseAnchorState uint8
+
+const (
+	// ForceCloseAnchorStateLimbo is set if the recovered_balance is zero
+	// and limbo_balance is non-zero.
+	ForceCloseAnchorStateLimbo ForceCloseAnchorState = iota
```
guggero, line 693: Hmm, this seems brittle if we use `iota` for types that we declare in the RPC. Sure, we most likely never change existing RPC values. But maybe we could still just do a shorter const alias here? e.g. `ForceCloseAnchorStateLimbo ForceCloseAnchorState = lnrpc.PendingChannelsResponse_ForceClosedChannel_LIMBO`?

guggero (approved):
LGTM :tada:

lightning_client.go
bhandras: Yeah agree. Changed to your suggestion.

PR #140, opened 2023-01-12
chainkit: fix macaroon

bhandras:
This commit removes the unneded chainkit.macaroon dependency as the chainkit RPC service currently uses the chainnoitifer.macaroon.

#### Pull Request Checklist


guggero (approved):
Ah, nice find.

PR #142, opened 2023-01-27
walletkit: add ListSweepsVerbose method to fetch verbose tx details

bhandras:
This commit adds a new method to the walletkit client to allow fetching sweeps with rich details as provided by LND.

#### Pull Request Checklist


lnd_services.go
```diff
@@ -42,7 +42,7 @@ var (
 	minimalCompatibleVersion = &verrpc.Version{
 		AppMajor:  0,
 		AppMinor:  15,
-		AppPatch:  4,
+		AppPatch:  5,
```
guggero, line 45: Do we need to bump the minimal compatible version here? It looks to me like we only need the compile time dependency for the output script types, but the RPC call should be available in earlier versions as well?

walletkit_client.go
```diff
@@ -424,6 +434,153 @@ func (m *walletKitClient) ListSweeps(ctx context.Context) ([]string, error) {
...
+		&walletrpc.ListSweepsRequest{
+			Verbose: true,
+		},
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	var result []lnwallet.TransactionDetail
+	// Since we have requested the abbreviated response from lnd, we can
```
guggero, line 566: nit: this comment doesn't match the actual code (seems to be copied from the non-verbose version).

walletkit_client.go
```diff
@@ -424,6 +434,153 @@ func (m *walletKitClient) ListSweeps(ctx context.Context) ([]string, error) {
...
+	// just get our response to a list of sweeps and return it.
+	rpcDetails := resp.GetTransactionDetails()
+	if rpcDetails != nil {
```
guggero, line 569: Should we error out if this is nil? Since we explicitly requested the verbose version.

guggero (commented):
Nice :tada:

lnd_services.go
bhandras: Yes indeed, kept it at 0.15.4.

walletkit_client.go
bhandras: Ah yes, thanks! Fixed.

guggero (approved):

PR #149, opened 2023-05-04
lightning: fix RegisterRPCMiddleware error handling

ellemouton:
Previously, the interceptStream's context was listened on for any context cancelation. However, this is not required since if the interceptStream's context is canceled then the `Recv()` method would also return a "context canceled" error which would be put into the `errChan` which is then listened on. If we listen on both the ctx.Done() channel and the error channel then the ctx channel will always fire first which results in a less meaningful error being returned.


guggero (approved):
Needs a rebase to fix CI, otherwise LGTM :tada:

PR #151, opened 2023-05-23
Regenerate macaroons when required permissions change

ViktorTigerstrom:
This PR addresses issue [#540](https://github.com/lightninglabs/lightning-terminal/issues/540) in the lightning-terminal repo.

With this PR, we regenerate macaroons if the required permissions in the service implementing the `lndclient.MacaroonService` have changed since last starting an `MacaroonService` instance.

One specific question I'd like some feedback on is:
I was a bit unsure if we'd like to regenerate the macaroon if we have removed one or more required permissions since last starting the service. Is there any scenario where we wouldn't like that to be the case? Two potential scenarios where I can see this specific functionality not being desired is:
1. The same service has uses two or more `MacaroonService` instances, which refer to the same macaroon, but one of the services only has a subset of the required permissions. If that was the case, the service would keep regenerating the macaroon for every `MacaroonService` instance. Do we believe that this is a scenario we'll ever have in practice?
2. Two or more services use the same macaroon, where one service only has a subset of the required permissions. Though this should never happen right, as that wouldn't make much sense?
I've added two fixup commits for the two main commits that implements that functionality, so let me know if you'd like me to fixup those commits or drop them.

Happy to address any feedback :)!


macaroon_service.go
```diff
@@ -367,3 +412,50 @@ func NewBoltMacaroonStore(dbPath, dbFileName string,
...
+	// Make sure it actually is a macaroon by parsing it.
+	oldMac := &macaroon.Macaroon{}
+	if err := oldMac.UnmarshalBinary(macBytes); err != nil {
+		return false, fmt.Errorf("unable to decode macaroon: %v", err)
+	}
+
+	macSlice := macaroon.Slice{oldMac}
+	authChecker := ms.Checker.Auth(macSlice)
+
+	ctx := context.Background()
```
ViktorTigerstrom, line 435: Would really appreciate some feedback on how to use contexts here:  
I still don't feel that I fully understand the best practices for how context(s) should be used. Is ok to create a new context here, or should this rather be done in the outer service that uses the lndclient, and then pass that context to the `MacaroonService` via the `Start` function?

macaroon_service.go
```diff
@@ -367,3 +410,37 @@ func NewBoltMacaroonStore(dbPath, dbFileName string,
...
+	requiredPerms := extractPerms(ms.cfg.RequiredPerms)
+
+	_, err = authChecker.Allow(ctx, requiredPerms...)
+	if err != nil {
+		// In case error here, it's most likely because the old macaroon
```
ellemouton, line 438: nit: s/In case error here/If an error is returned here

macaroon_service.go
```diff
@@ -441,6 +441,19 @@ func (ms *MacaroonService) macaroonMatchesPermissions() (bool, error) {
 		return false, nil
 	}
 
+	// If the number of ops in the allowed info is not the same as the
+	// number of required permissions, ie. there's less required permissions
+	// than allowed ops, then the required permissions have been modified to
+	// require less permissions than the old macaroon has. We therefore
```
ellemouton, line 447: s/less/fewer

macaroon_service_test.go
```diff
@@ -94,6 +96,122 @@ func TestMacaroonServiceMigration(t *testing.T) {
 	require.NoError(t, ms.Stop())
 }
 
+// TestMacaroonGeneration tests that the macaroon service generates macaroons
```
ellemouton, line 99: suggest:

```
// TestMacaroonGeneration tests that the macaroon service generates macaroons
// as expected. A macaroon should not be regenerated if the required permissions
// haven't changed, and should be regenerated if the required permissions have 
// changed.
```

macaroon_service_test.go
```diff
@@ -94,6 +96,122 @@ func TestMacaroonServiceMigration(t *testing.T) {
...
+// properly, and doesn't regenerate the macaroon if the required permissions
+// haven't changed, but that it also does regenerate the macaroon when required
+// as the required permissions change.
+func TestMacaroonGeneration(t *testing.T) {
+	// Create a temporary directory where we can store the macaroon db and
+	// the macaroon we are about to create.
+	tempDirPath, err := ioutil.TempDir("", ".testMacaroons")
+	require.NoError(t, err)
+	defer os.RemoveAll(tempDirPath)
+
```
ellemouton, line 109: can replace the whole thing with:

`tempDirPath := t.TempDir()`

macaroon_service_test.go
```diff
@@ -94,6 +96,122 @@ func TestMacaroonServiceMigration(t *testing.T) {
...
+	// Set the macaroon store, and get the root key store.
+	rks, backend, err := NewBoltMacaroonStore(
+		tempDirPath, "macaroons.db", defaultDBTimeout,
+	)
+	require.NoError(t, err)
+	defer func() {
+		require.NoError(t, backend.Close())
+	}()
```
ellemouton, line 117:
```
t.Cleanup(func() {
		require.NoError(t, backend.Close())
})
```

macaroon_service_test.go
```diff
@@ -94,6 +96,122 @@ func TestMacaroonServiceMigration(t *testing.T) {
...
+	firstEntityRead := bakery.Op{Entity: "first_entity", Action: "read"}
+	secondEntityRead := bakery.Op{Entity: "second_entity", Action: "read"}
+
+	mockPerms := make(map[string][]bakery.Op)
+
+	// 1. We start by adding 2 permissions to the mockPerms map, and ensure
+	// that the macaroon service generates a macaroon with the correct
+	// permissions.
+	mockPerms["/uri1.Test/WriteTest"] = []bakery.Op{firstEntityWrite}
+	mockPerms["/uri1.Test/ReadTest"] = []bakery.Op{firstEntityRead}
```
ellemouton, line 141:
```
	mockPerms := map[string][]bakery.Op{
		"/uri1.Test/WriteTest": {firstEntityWrite},
		"/uri1.Test/ReadTest":  {firstEntityRead},
	}
```

macaroon_service_test.go
```diff
@@ -147,3 +265,41 @@ func (m *mockSignerClient) DeriveSharedKey(_ context.Context,
...
+	require.NoError(t, err)
+
+	mac := &macaroon.Macaroon{}
+	err = mac.UnmarshalBinary(macBytes)
+	require.NoError(t, err)
+
+	return mac, macBytes
+}
+
+func ensureMacaroonPermissions(t *testing.T, mac *macaroon.Macaroon,
```
ellemouton, line 306: suggestion: `assertMacaroonPerms`

macaroon_service_test.go
```diff
@@ -147,3 +265,41 @@ func (m *mockSignerClient) DeriveSharedKey(_ context.Context,
...
+	expectedPermissions map[string][]bakery.Op, ms *MacaroonService) {
+
+	macSlice := macaroon.Slice{mac}
+	authChecker := ms.Checker.Auth(macSlice)
```
ellemouton, line 310: nit: can in-line

macaroon_service_test.go
```diff
@@ -94,6 +96,122 @@ func TestMacaroonServiceMigration(t *testing.T) {
...
+	// correct permissions.
+	mac1, mac1Bytes := extractMacaroon(t, cfg.MacaroonPath)
+	ensureMacaroonPermissions(t, mac1, mockPerms, ms)
+
+	require.NoError(t, ms.Stop())
+
+	// 2. Secondly, we create a new macaroon service with the same
+	// permissions as the first macaroon service, and ensure that the
+	// macaroon hasn't been regenerated.
+	ms = generateMacaroonService(t, cfg)
```
ellemouton, line 156: you can just keep stopping and starting the same macaroon service. No need to keep generating a new one.

macaroon_service_test.go
```diff
@@ -94,6 +96,122 @@ func TestMacaroonServiceMigration(t *testing.T) {
...
+	// macaroon, as if it did, the macaroon bytes would be different as
+	// another nonce would have been used when generating mac2, resulting
+	// in different macaroon bytes (the signature for the macaroon would
+	// also differ).
+	require.Equal(t, mac1Bytes, mac2Bytes)
+	require.NoError(t, ms.Stop())
+
+	// 3. Thirdly, we create a new macaroon service, but this time we add a
+	// permission to the mockPerms map that is for a different URI, but that
+	// requires the an entity and action we already have in the previous
```
ellemouton, line 170: nit: "the an"

macaroon_service_test.go
```diff
@@ -94,6 +96,122 @@ func TestMacaroonServiceMigration(t *testing.T) {
...
+	ms = generateMacaroonService(t, cfg)
+
+	_, mac3Bytes := extractMacaroon(t, cfg.MacaroonPath)
+
+	require.Equal(t, mac1Bytes, mac3Bytes)
+	require.NoError(t, ms.Stop())
+
+	// 4. Fourthly, we create a new macaroon service, but this time we add a
+	// permission to the mockPerms map that requires a different entity and
+	// action than the we already have in the previous macaroon. We then
```
ellemouton, line 185: nit: "than the we"

macaroon_service_test.go
```diff
@@ -94,6 +96,122 @@ func TestMacaroonServiceMigration(t *testing.T) {
 	require.NoError(t, ms.Stop())
 }
 
+// TestMacaroonGeneration tests that the macaroon service generates macaroons
+// properly, and doesn't regenerate the macaroon if the required permissions
+// haven't changed, but that it also does regenerate the macaroon when required
+// as the required permissions change.
```
ellemouton, line 102: Excellent test!!! 🚀

macaroon_service.go
ellemouton: This should be fine as is cause I dont think `.Allow` is at all expected to hang. Yes it is fine to create a new context here. What you could do to be safe if you wanted is:

```
	ctx, cancel := context.WithTimeout(context.Background(), time.Second*5)
	_, err = authChecker.Allow(ctx, requiredPerms...)
	cancel()
```
(and then just put the `time.Second * 5` part into a const).

macaroon_service.go
```diff
@@ -367,3 +410,50 @@ func NewBoltMacaroonStore(dbPath, dbFileName string,
...
+	}
+
+	// Make sure it actually is a macaroon by parsing it.
+	oldMac := &macaroon.Macaroon{}
+	if err := oldMac.UnmarshalBinary(macBytes); err != nil {
+		return false, fmt.Errorf("unable to decode macaroon: %v", err)
+	}
+
+	macSlice := macaroon.Slice{oldMac}
+	authChecker := ms.Checker.Auth(macSlice)
```
ellemouton, line 431: can in-line :) 

```
var (
		authChecker   = ms.Checker.Auth(macaroon.Slice{oldMac})
		requiredPerms = extractPerms(ms.cfg.RequiredPerms)
)
```

ellemouton (commented):
Excellent!

macaroon_service.go
```diff
@@ -231,12 +233,53 @@ func (ms *MacaroonService) Start() error {
...
+		if matches {
+			return nil
+		}
+
+		if err != nil {
+			// In case we error while checking if the old macaroon
+			// permissions match the required permissions, we still
+			// go ahead and create a new macaroon, but make sure to
+			// log the error.
+			log.Warnf("Unable to match the old macaroon "+
```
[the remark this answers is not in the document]
ViktorTigerstrom: Oh yeah, when I think about it, I agree that it's better to return the error than just blindly overwrite the path. My previous idea was that a user would likely always want the macaroon generation to succeed if it was possible to make it succeed. My idea was that if we'd error parsing the macaroon there would be some corruption or similar in the macaroon that the user would likely want the service to fix if possible, but it makes sense that a more likely scenario would be that the path is not actually pointing to a macaroon, which in that case we wouldn't like to overwrite.

macaroon_service_test.go
ViktorTigerstrom: Ah neat! Thanks for that :)!

macaroon_service_test.go
ViktorTigerstrom: Thanks! 🚀

macaroon_service.go
ViktorTigerstrom: Thanks a lot for explanation and for the suggested safety fix! I've implemented it with the latest push :)

ViktorTigerstrom:
Thanks a lot for the review and the great feedback @ellemouton!! I've addressed your feedback with the latest push :)

macaroon_service.go
```diff
@@ -231,12 +234,46 @@ func (ms *MacaroonService) Start() error {
...
+		log.Infof("Macaroon at %s does not have all required "+
+			"permissions. Deleting it and creating a new "+
+			"one", ms.cfg.MacaroonPath)
+
+		if err := os.Remove(ms.cfg.MacaroonPath); err != nil {
+			return fmt.Errorf("Unable to remove macaroon "+
+				"at %s: %v", ms.cfg.MacaroonPath, err)
+		}
+	}
+
```
ellemouton, line 276: also , while we are here, do you mind replacing the `ioutil.WriteFile` below with `os.WriteFile`?

macaroon_service.go
```diff
@@ -231,12 +234,46 @@ func (ms *MacaroonService) Start() error {
...
+			return nil
+		}
+
+		// Else if the required permissions have been updated, we delete
+		// the old macaroon and create a new one.
+		log.Infof("Macaroon at %s does not have all required "+
+			"permissions. Deleting it and creating a new "+
+			"one", ms.cfg.MacaroonPath)
+
+		if err := os.Remove(ms.cfg.MacaroonPath); err != nil {
```
ellemouton, line 271: I think perhaps better not to remove the file and just allow `os.WriteFile` below to overwrite the file (it does that by default). Otherwise there is an intermediary state here where perhaps other services making use of this macaroon for calls all of a sudden cant find it.

macaroon_service.go
```diff
@@ -231,12 +234,46 @@ func (ms *MacaroonService) Start() error {
...
+			return fmt.Errorf("Unable to remove macaroon "+
+				"at %s: %v", ms.cfg.MacaroonPath, err)
+		}
+	}
+
 	// We don't offer the ability to rotate macaroon root keys yet, so just
 	// use the default one since the service expects some value to be set.
```
ellemouton, line 278: also, below on line 301 where we remove the file if writing to it failed: this was ok before because we would only get to that part if the file did not exist in the first place. But now, since we might be overwriting a file that did already exist, we should not remove it in that case if we fail

ellemouton (commented):
tACK 🔥  almost there!

macaroon_service.go
ellemouton: no i mean another service reading the macaroon to make calls. Not writing to it

macaroon_service.go
ViktorTigerstrom: I pushed this as a separate fixup commit right now, as I'd like to get your feedback on it before squashing it:
I see a risk if we remove this file removal code, that the Write code might fail during the execution for various reasons, corrupting the macaroon file. If that would happen, we'd now end up in a state where the permission comparison code will fail to parse it to a macaroon, and we'd never write a new macaroon file without manual removal of the file?

macaroon_service.go
ViktorTigerstrom: Oh yeah makes sense, thanks for the clarification!

macaroon_service.go
ellemouton: i think that is ok - ie to force a manual removal of the file in these rare cases. I'd rather the user make this decision than assume that we can remove it

ViktorTigerstrom: Ok sounds good :)! Squashing the commit then.

ViktorTigerstrom: Squashed with the latest push.

ellemouton (approved):
LGTM pending the squash of the commits 🔥  I will scout for a second reviewer who we can get a second opinion from regarding the regen of the mac for the cases where the set decreases

macaroon_service.go
```diff
@@ -231,12 +233,53 @@ func (ms *MacaroonService) Start() error {
...
+		return nil
+	}
+
+	// If we are not in stateless mode and a macaroon file does exist, we
+	// check that the macaroon matches the required permissions. If not, we
+	// will delete the macaroon and create a new one.
+	if lnrpc.FileExists(ms.cfg.MacaroonPath) {
+		matches, err := ms.macaroonMatchesPermissions()
+		if err != nil {
+			log.Warnf("An error occurred when attempting match "+
```
guggero, line 253: nit: s/attempting match/attempting to match/

macaroon_service.go
```diff
@@ -231,12 +233,53 @@ func (ms *MacaroonService) Start() error {
...
-	if ms.cfg.StatelessInit || lnrpc.FileExists(ms.cfg.MacaroonPath) {
+	// mode).
+	if ms.cfg.StatelessInit {
 		return nil
 	}
 
+	// If there are no required permissions, then we don't have to generate
+	// or re-generate a macaroon and so we can exit here.
+	if len(ms.cfg.RequiredPerms) == 0 {
+		return nil
```
guggero, line 244: I think we should return an error here instead. A macaroon without any permissions doesn't really serve any purpose and is potentially a security risk. So I think it makes sense to move this check to the `NewMacaroonService` function.

guggero (approved):
Very nice, LGTM :tada: 

I checked that the assumptions we make in the macaroon service about which fields are set in the `MacaroonServiceConfig` are correct in our projects (loop, pool, faraday, taproot-assets, litd), specifically that `MacaroonPath` and `RequiredPerms` are always set.

macaroon_service.go
ViktorTigerstrom: Hmm so we have another test in the lndclient repo `TestMacaroonServiceMigration` that will fail if we return an error here. Since that test isn't creating any macaroon `RequiredPerms` will be `nil`. 
So if we're not ok with returning `nil` if this if clause is true, I see two other alternatives:
1. Change that test to also create a Macaroon, but that would kind of go against the focus of the test.
2. Instead of matching using the `RequiredPerms`, match by checking the `cfg.MacaroonPath` and checking if the path ends with `.macaroon`, and then return nil. Not really a fan of that alternative either as it will be a bit confusing if people integrate the `lndclient` repo that the repo will return `nil` and not an error if the macaroon path if incorrect.

Which version would you prefer :)?

ViktorTigerstrom:
Thanks a lot for the review @guggero!! Let me know how we'd like to tackle https://github.com/lightninglabs/lndclient/pull/151#discussion_r1206762513 and I'll update with that change :)!

Also let me know if you think it's ok to squash the commits, or if we shouldn't regenerate the macaroon if the number of required permissions have been reduced (i.e. the question in https://github.com/lightninglabs/lndclient/pull/151#issue-1722937957).

macaroon_service.go
guggero: If I interpret the diff correctly, I think `TestMacaroonServiceMigration` did create a macaroon before, just with empty permissions (and it didn't explicitly check for the creation).
So I think option 1 is fine, just adding a dummy permission to that test (1 line change).

guggero:
> Also let me know if you think it's ok to squash the commits

Yes, I think those fixup commits are good, you can squash them.

macaroon_service.go
ViktorTigerstrom: Ok thanks! I added generation of a macaroon to the `TestMacaroonServiceMigration` test :)! The previous diff succeeded without the macaroon as the code would exit early since the `tempDirPath` existed and `lnrpc.FileExists(ms.cfg.MacaroonPath)` would then match.

ViktorTigerstrom:
Just pushed change that errors if we have no required permission, and also added extra test coverage of the things we've updated the PR with all the feedback :)!

ViktorTigerstrom:
Snap sorry, managed to get the addition of the error into the second commit. Give me 2min and I'll fix that.

macaroon_service_test.go
```diff
@@ -94,6 +100,159 @@ func TestMacaroonServiceMigration(t *testing.T) {
...
+
+	// We add 2 required permissions to the mockPerms map, which would make
+	// the previous start attempts succeed if we start the service now.
+	mockPerms["/uri1.Test/WriteTest"] = []bakery.Op{firstEntityWrite}
+	mockPerms["/uri1.Test/ReadTest"] = []bakery.Op{firstEntityRead}
+	cfg.RequiredPerms = mockPerms
+
+	// But we also test that we fail to start the macaroon service if we set
+	// the MacaroonPath to an existing file which isn't a macaroon.
+	cfg.MacaroonPath = filepath.Join(tempDirPath, "macaroons.db")
```
ViktorTigerstrom, line 162: Just wanted to check:
If we set the macaroon path a non existing file with another file extension than .macaroon, we'll still generate the macaroon, and write the file to that path.
This is expected behavior correct, as it's technically a macaroon file even though it has another file extension?

ellemouton: yeah I dont think the extension matters :) I think the behaviour is expected :)

ViktorTigerstrom: Awesome thanks!

macaroon_service.go
```diff
@@ -231,12 +233,54 @@ func (ms *MacaroonService) Start() error {
...
+	// mode).
+	if ms.cfg.StatelessInit {
 		return nil
 	}
 
+	// If there are no required permissions, then we can't create a macaroon
+	// and we will return an error.
+	if len(ms.cfg.RequiredPerms) == 0 {
+		return fmt.Errorf("the config field RequiredPerms must to be " +
+			"set and must contain elements to create a macaroon")
```
ellemouton, line 245: I think Oli was saying we should move this check into `NewMacaroonService` right?

ViktorTigerstrom: Oh sorry, missed that part. Will fix this!

PR #152, opened 2023-06-04
multi: hanlde empty invoice preimages for AMP invocies

positiveblue:
An AMP invoice does not have an invoice-level preimage. Each htlc part of an AMP payment has its own hash/preimage pair.

This was blocking my litd from starting up. The account service subscribes to the invoice stream and without this changes lndclient will return an error when it is not able to parse the preimage for a settled invoice.

```
2023-05-29 19:20:24.075 [ERR] ACCT: Error in invoice subscription: invalid preimage length of 0, want 32
```

#### Pull Request Checklist

### DIFF
--- a/lightning_client.go
+++ b/lightning_client.go
@@ -1483,12 +1483,21 @@ func (s *lightningClient) payInvoice(ctx context.Context, invoice string,
 				)
 
 				r := payResp.PaymentRoute
-				preimage, err := lntypes.MakePreimage(
-					payResp.PaymentPreimage,
+
+				var (
+					preimage lntypes.Preimage
+					err      error
 				)
-				if err != nil {
-					return &PaymentResult{Err: err}
+
+				if payResp.PaymentPreimage != nil {
+					preimage, err = lntypes.MakePreimage(
+						payResp.PaymentPreimage,
+					)
+					if err != nil {
+						return &PaymentResult{Err: err}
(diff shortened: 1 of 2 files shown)


lightning_client.go
```diff
@@ -1483,12 +1483,21 @@ func (s *lightningClient) payInvoice(ctx context.Context, invoice string,
...
-					payResp.PaymentPreimage,
+
+				var (
+					preimage lntypes.Preimage
+					err      error
 				)
-				if err != nil {
-					return &PaymentResult{Err: err}
+
+				if payResp.PaymentPreimage != nil {
```
ellemouton, line 1492: should we maybe first check if this is actually an AMP invoice by checking `payReq.Features.IsSet(lnwire.AMPRequired)`? and then we only allow the pre-image to be nil if this is actually an AMP payment? 

perhaps that is overkill... lemme know what you think :)

ellemouton (approved):
LGTM 🔥  great catch. Left one question about possibly first checking if the invoice has the AMP feature bit set before allowing the pre-image to be nil

guggero (approved):
LGTM :tada:

lightning_client.go
guggero: Discussed offline, want to avoid having as much logic in `lndclient` as possible.

PR #158, opened 2023-09-22
mod+lightning: update to lnd v0.17.0-beta.rc4

guggero:
I created a new branch `lnd-17-0` (which currently is equal to `0-16-0`), this is the first PR on it.

We simply bumpy the lnd version to `v0.17.0-beta.rc4` and add the commitment type for Simple Taproot channels.

#### Pull Request Checklist


bhandras (approved):
LGTM 🚀

PR #167, opened 2024-01-01
invoices_client: account for invoice hash being nil.

ardevd:
When the invoice hash is nil, a random one should be generated for us. Hence, the `AddHoldInvoice` function must account for a nil value.


ardevd:
Nevermind, I realize `AddHoldInvoice` is tied to the supplied hash, so one must be provided, if I understand the logic correctly.

PR #171, opened 2024-01-14
client: implemented messaging signing.

ardevd:
added support for the main RPC implementation of message signing.

### DIFF
--- a/lightning_client.go
+++ b/lightning_client.go
@@ -271,6 +271,13 @@ type LightningClient interface {
 	// relevant to the wallet are sent over.
 	SubscribeTransactions(ctx context.Context) (<-chan Transaction,
 		<-chan error, error)
+
+	// SignMessage signs a message with this node's private key.
+	// The returned signature string is zbase32 encoded and pubkey
+	// recoverable, meaning that only the message digest and signature
+	// are needed for verification.
+	SignMessage(ctx context.Context, data []byte) (string,
+		error)
 }
 
 // Info contains info about the connected lnd node.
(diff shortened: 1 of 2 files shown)


lightning_client.go
```diff
@@ -4132,6 +4139,20 @@ func (s *lightningClient) SendCustomMessage(ctx context.Context,
 	return err
 }
 
+// SignMessage signs a message with this node's private key.
+// The returned signature string is zbase32 encoded and pubkey
+// recoverable, meaning that only the message digest and signature
+// are needed for verification.
+func (s *lightningClient) SignMessage(ctx context.Context,
+	data []byte) (string, error) {
+	rpcCtx := s.adminMac.WithMacaroonAuth(ctx)
```
guggero, line 4148: nit: extra line between multi-line function header and first code.

lightning_client.go
```diff
@@ -4132,6 +4139,20 @@ func (s *lightningClient) SendCustomMessage(ctx context.Context,
...
+
+	rpcReq := &lnrpc.SignMessageRequest{Msg: data}
+	rpcRes, err := s.client.SignMessage(rpcCtx, rpcReq)
```
guggero, line 4151: Need to add an `if err != nil ` check here. Otherwise `rpcRes.Signature` below will cause a nil pointer panic.

guggero (changes requested):
Thanks for the PR!

lightning_client.go
ardevd: Ah, nice catch. Should have caught that one.

lightning_client.go
```diff
@@ -4145,10 +4145,13 @@ func (s *lightningClient) SendCustomMessage(ctx context.Context,
...
-	rpcCtx := s.adminMac.WithMacaroonAuth(ctx)
 
+	rpcCtx := s.adminMac.WithMacaroonAuth(ctx)
 	rpcReq := &lnrpc.SignMessageRequest{Msg: data}
 	rpcRes, err := s.client.SignMessage(rpcCtx, rpcReq)
+	if err != nil {
+		return "", err
+	}
 
 	return rpcRes.Signature, err
```
guggero, line 4156: Should now just return `rpcRes.Signature, nil` here.

ardevd: Sharp eyed reviewer! Fixed.

bhandras (approved):
I believe you need to update `permissions.json` to make the test pass.
Other than that LGTM 🎉

ardevd:
> I believe you need to update `permissions.json` to make the test pass. Other than that LGTM 🎉

Not entirely sure what update is needed to `permissions.json`. Something like:

```json
"/lnrpc.Lightning/Signmessage": {
            "permissions": [
                {
                    "entity": "onchain",
                    "action": "read"
                }
            ]
        },
```
?

guggero:
Yes, that should be enough.

ardevd:
Actually, on further inspection, a similar block is already present in permissions.json.

```json
"/lnrpc.Lightning/SignMessage": {
            "permissions": [
                {
                    "entity": "message",
                    "action": "write"
                }
            ]
        },
```

on line 563

guggero:
Ah, I think you need to bump Line 15 in https://github.com/lightninglabs/lndclient/blob/2005b0d5f6a33397a7679bb722ed096c77fe55c2/macaroon_recipes_test.go#L15.
Since a new permission is now being used by the `lndclient`.

PR #172, opened 2024-01-15
refactor: replaced deprecated ioutil calls with os equivalents

ardevd:


basic_client_test.go
```diff
@@ -46,21 +45,21 @@ XhkpT5dliEGFLNe6OOgeWTU1JpEXfCud/GImtNMHQi4EDWQfvWuCNGhOoQ==
 
 	// Now let's write the data to a file to make sure parseTLSAndMacaroon
 	// parses that properly as well.
-	tempDirPath, err := ioutil.TempDir("", ".testCreds")
+	tempDirPath, err := os.MkdirTemp("", ".testCreds")
```
guggero, line 48: Can even use `t.TempDir()`, then the `defer os.RemoveAll(tempDirPath)` is done automatically.

guggero (approved):
LGTM, thanks!

bhandras (approved):
LGTM, thanks! 🌮

basic_client_test.go
ardevd: Nice! Agree!

guggero:
Linter is not happy because of the extra space.

ardevd:
Fixed!

PR #178, opened 2024-03-06
lndclient: add rpc/daemon state client to services

calvinrzachman:
## Objective

Add the `state` RPC client to the list of exported LND services. This may be useful in assessing health of `lnd` instances from the payment service's health checking routines.

#### Pull Request Checklist


guggero (approved):
Nice catch! LGTM :tada:

PR #185, opened 2024-06-03
go.mod: update LND to v0.18.0-beta.1

starius:
This is the same as v0.18.0-beta.

#### Pull Request Checklist


bhandras (approved):
LGTM 💯

tools/Dockerfile
```diff
@@ -11,7 +11,7 @@ RUN cd /tmp \
   && mkdir -p /tmp/build/.cache \
   && mkdir -p /tmp/build/.modcache \
   && cd /tmp/tools \
-  && go install -trimpath github.com/golangci/golangci-lint/cmd/golangci-lint \
+  && go install -trimpath github.com/golangci/golangci-lint/cmd/golangci-lint@v1.59.0 \
```
guggero, line 14: This should be updated in the `go.mod` file in the `tools` folder. That's the whole point of that `go.mod` file in the first place, that we can pin version numbers of our tooling binaries we want to use.

guggero (commented):
Thanks a lot for the fixes! One request, otherwise looks good :tada:

PR #190, opened 2024-07-24
Add `RawClientWithMacAuth` to each client

guggero:
This allows us to use the underlying RPC methods directly while still taking advantage of the macaroon management.

Using the raw RPC methods directly can be useful when a new method isn't implemented in `lndclient` yet or when we want to pass the raw RPC request method directly instead of converting it to the `lndclient` structs.

#### Pull Request Checklist

### DIFF
--- a/chainkit_client.go
+++ b/chainkit_client.go
@@ -14,6 +14,8 @@ import (
 
 // ChainKitClient exposes chain functionality.
 type ChainKitClient interface {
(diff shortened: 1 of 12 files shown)


bhandras (approved):
LGTM, uTACK 🎉

ellemouton (approved):
lgtm!

PR #197, opened 2024-10-22
go.mod: update btclog dep

ellemouton:
For structured logging abilities.

I guess we'll want to wait for 0.19 branch? Pushing the PR so long so I can point to it for the update in other repos.


go.mod
```diff
@@ -6,10 +6,10 @@ require (
...
 	github.com/btcsuite/btcd/btcutil/psbt v1.1.8
 	github.com/btcsuite/btcd/chaincfg/chainhash v1.1.0
-	github.com/btcsuite/btclog v0.0.0-20170628155309-84c8d2346e9f
-	github.com/btcsuite/btcwallet v0.16.10-0.20240809133323-7d3434c65ae2
-	github.com/btcsuite/btcwallet/wtxmgr v1.5.3
-	github.com/lightningnetwork/lnd v0.18.3-beta.rc3.0.20241011124628-ca3bde901eb8
+	github.com/btcsuite/btclog/v2 v2.0.0-20241017175713-3428138b75c7
+	github.com/btcsuite/btcwallet v0.16.10-0.20240912233857-ffb143c77cc5
+	github.com/btcsuite/btcwallet/wtxmgr v1.5.4
+	github.com/lightningnetwork/lnd v0.18.0-beta.rc4.0.20241022180056-b4ef22bf4722
```
ellemouton, line 12: unsure of why it reverts back to 18.0. Thought it would build on-top of 18.3 🤔 ie, thought it would take the latest tag on master and then point to the commit in question which comes after that tag

ellemouton: ahhhh - I guess v0.18.3 is  actually not on master. My bad. 

So I assume we will want to wait for a 0.19 tag  there?

guggero: Don't worry about the beginning of the pseudo version, the commit is the important part. And yes, 0.18.3 isn't in master, so v0.18.0 as the prefix is to be expected.
We can wait but IMO we don't have to. That's why we have the different branches in `lndclient`, so we can actually start integrating future releases, such as the not-yet-fully-0.19.0 version.

ellemouton: ok cool - so this is g2g in your opinion? 
(once we get this in, the other deps can start being updated)

guggero: Yes, this is good to go. But I think we should wait with any of the other dependencies, otherwise it will interfere with the `lnd 0.18.4` update that is expected to come soon. I mean we can create all the PRs to have them queued up, but we should only actually merge them close to 0.19 being out.

guggero (approved):

bhandras:
I think this now needs rebase.

PR #198, opened 2024-10-24
invoiecs: fix value conversion bug in AddHodlInvoice

Roasbeef:
Before this commit, callers wouldn't be able to use an msat amount as input to the hold invoice RPC call. Instead, the wrapper logic would always convert to msat first, causing a loss in precision. For most cases, sat amounts are fine, but some use cases require higher precision.

We fix this by keeping the input value in mSat (as typed), while using the `ValueMsat` field as `lnd` would expect.

#### Pull Request Checklist


bhandras (approved):
Nice catch! LGTM 💯

guggero (approved):
Nice catch! LGTM :tada:

guggero:
I'm also going to cherry pick that comment onto the master branch, since it's a useful fix there too.

PR #200, opened 2024-10-29
Minrelayfee func

gijswijs:
Added MinRelayFee func to walletkit_client.go.

#### Pull Request Checklist

### DIFF
--- a/walletkit_client.go
+++ b/walletkit_client.go
@@ -111,6 +111,10 @@ type WalletKitClient interface {
 	EstimateFeeRate(ctx context.Context,
 		confTarget int32) (chainfee.SatPerKWeight, error)
 
+	// MinRelayFee returns the current minimum relay fee based on our chain
+	// backend in sat/kw.
+	MinRelayFee(ctx context.Context) (chainfee.SatPerKWeight, error)
+
 	// ListSweeps returns a list of sweep transaction ids known to our node.
 	// Note that this function only looks up transaction ids, and does not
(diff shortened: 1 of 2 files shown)


bhandras (approved):
LGTM 🌮

walletkit_client.go
```diff
@@ -540,6 +540,25 @@ func (m *walletKitClient) EstimateFeeRate(ctx context.Context, confTarget int32)
 	return chainfee.SatPerKWeight(resp.SatPerKw), nil
 }
 
+// MinRelayFee returns the current minimum relay fee based on our chain backend
+// in sat/kw.
+func (m *walletKitClient) MinRelayFee(ctx context.Context) (
```
guggero, line 545: nit: I would've formatted this as:

```
func (m *walletKitClient) MinRelayFee(
	ctx context.Context) (chainfee.SatPerKWeight, error) {
```

But non-blocking.

guggero (approved):
LGTM :tada: 

I though we could do this against the `master` branch, but it turns out the new field was actually only added in `0.18.x`, so `lnd-18-4` is the correct branch.

guggero:
Can you squash the commits? Then I think we can merge this. Or is anything else missing here? Just waiting to confirm everything works in `tapd`?

gijswijs:
> Just waiting to confirm everything works in `tapd`?
Exactly. All seems to work. So we have a cascade of PRs depending on this one.

PR #203, opened 2024-11-22
signer: create workaround for SignOutputRaw quirk

guggero:
This commit fixes a long-standing issue with how SignOutputRaw populates the key descriptor before calling into lnd.
If the public key is available, _only_ the public key is populated and the key locator (index+family) is not. That works well for any keys the wallet is aware of.
But if a wallet is restored from seed, it will not know any addresses/keys apart from index 0 of each family/account. And a lookup by public key only will fail.
To fix that, we add a new method SignOutputRawKeyLocator that has an updated behavior that also sends along the key locator if we're certain it is fully known.

Because changing any behavior in this area of the code might lead to breaking existing behavior some clients like Loop or Pool rely on, we explicitly don't change the original method but rather add a new one.

#### Pull Request Checklist


guggero:
Will help solve the problem described in https://github.com/lightninglabs/taproot-assets/issues/1208.

jharveyb (approved):
LGTM 🎉

Roasbeef:
> But if a wallet is restored from seed, it will not know any addresses/keys apart from index 0 of each family/account. And a lookup by public key only will fail.

Correct that the public key look up will fail, however this should be accounted for with the existing logic: https://github.com/lightningnetwork/lnd/blob/94f7ed46deab346cc19ff1e6aa00cecdfe3eb775/keychain/btcwallet.go#L324-L363

This path is taken when we need to derive the private key based on the keyloc or keydesc. 

Checking out the diff now.

signer_client.go
```diff
@@ -215,26 +225,70 @@ func (s *signerClient) RawClientWithMacAuth(
...
+	// but the index might be zero because it's the first key, or because it
+	// isn't known at that particular moment.
+	// We aim to be compatible with this method in lnd's wallet:
+	// https://github.com/lightningnetwork/lnd/blob/master/lnwallet/btcwallet/signer.go#L286
+	// Because we know all custom families (0 to 255) are derived at wallet
+	// creation, and the very first index of each family/account is always
+	// derived, we know that only using the public key for that very first
+	// index will work. But for a freshly initialized wallet (e.g. restored
+	// from seed), we won't know any indexes greater than 0, so we _need_ to
+	// also specify the key locator and not just the public key.
```
Roasbeef, line 266: This is where I'm confused. If we only specify the public key, then we'll derive all private keys from 0 to 10k for that key family, checking the public key each time. 

With the way the logic works, there're 3 main paths: 
  1. You specify no public key, it checks the cache: https://github.com/lightningnetwork/lnd/blob/94f7ed46deab346cc19ff1e6aa00cecdfe3eb775/keychain/btcwallet.go#L266-L280
  2. Index is set, or no public key, it tries to derive from scratch: https://github.com/lightningnetwork/lnd/blob/94f7ed46deab346cc19ff1e6aa00cecdfe3eb775/keychain/btcwallet.go#L300-L322
  3. Only pubkey, it scans: https://github.com/lightningnetwork/lnd/blob/94f7ed46deab346cc19ff1e6aa00cecdfe3eb775/keychain/btcwallet.go#L324-L363

When we hit disk it eventually calls into `deriveKeyFromPath`: https://github.com/btcsuite/btcwallet/blob/66a3aeef6e78751e644a3d03fd856e982e0db4ac/waddrmgr/scoped_manager.go#L742-L769. DIsk wise, this only need the account to exist, as then it'll decrypt the master key and derive the addr as normal: https://github.com/btcsuite/btcwallet/blob/66a3aeef6e78751e644a3d03fd856e982e0db4ac/waddrmgr/scoped_manager.go#L385-L416

signer_client.go
```diff
@@ -215,26 +225,70 @@ func (s *signerClient) RawClientWithMacAuth(
...
+	fullDescriptor := func(
```
Roasbeef, line 267: What is resolved by setting _both_? From my reading of the code, only one or the other is read.

signer_client.go
guggero: We don't land in any of the code paths you linked... The problem is that the `signrpc.SignOutputRaw` RPC uses [`input.Signer.SignOutputRaw`](https://github.com/lightningnetwork/lnd/blob/789c6bac8ca4b75d15e9e40f8ded8e394df518bb/lnrpc/signrpc/signer_server.go#L477) which is implemented by [`BtcWallet.SignOutputRaw`](https://github.com/lightningnetwork/lnd/blob/b17db4a32ac370684f6c8a8e58ec89149cb2b3a9/lnwallet/btcwallet/signer.go#L357) which then calls into [`BtcWallet.fetchPrivKey`](https://github.com/lightningnetwork/lnd/blob/b17db4a32ac370684f6c8a8e58ec89149cb2b3a9/lnwallet/btcwallet/signer.go#L286) which has a completely different logic.

signer_client.go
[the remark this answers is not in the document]
guggero: I wanted to make this as clear and obvious as possible, with as much comment on _why_ we're doing it that way, so I went with two closures on purpose. Just so nobody else spends debug hours on this (because this is already the second or third time I ran into this, once in Loop and once in Pool).

PR #204, opened 2024-12-12
Expose all params for `AddHoldInvoice`

GeorgeTsagk:
## Description

Previously we'd only pass through a very limited set of the parameters to the LND RPC, and we'd hard-code certain values like `Private: true`.

With this PR we're forwarding the request as it was given to lndclient, without changing any of its body.

### DIFF
--- a/invoices_client.go
+++ b/invoices_client.go
@@ -3,6 +3,7 @@ package lndclient
 import (
 	"context"
 	"errors"
+	"fmt"
 	"io"
 	"sync"
 	"time"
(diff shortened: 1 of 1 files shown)


invoices_client.go
```diff
@@ -228,13 +230,43 @@ func (s *invoicesClient) AddHoldInvoice(ctx context.Context,
 	rpcCtx, cancel := context.WithTimeout(ctx, s.timeout)
 	defer cancel()
 
+	var routeHints []*lnrpc.RouteHint
```
bhandras, line 233: nit: could be preallocated.

bhandras (approved):
LGTM 🌮

Roasbeef (approved):
LGTM 🎤

invoices_client.go
```diff
@@ -228,13 +230,43 @@ func (s *invoicesClient) AddHoldInvoice(ctx context.Context,
...
+
+	for _, rh := range in.RouteHints {
```
guggero, line 235: Why not `routeHints, err := marshallRouteHints(in.RouteHints)`?

bhandras: Oh indeed I also didn't see that!

GeorgeTsagk: Oh, did not see this comment!
Much better, changing

PR #206, opened 2025-02-11
bumpfee: add more flags as options 

starius:
Use new fields instead of deprecated

'force' -> 'immediate'
'sat_per_byte' -> 'sat_per_vbyte'

New options added: 'immediate', 'target_conf', 'budget'.

#### Pull Request Checklist

### DIFF
--- a/walletkit_client.go
+++ b/walletkit_client.go
@@ -134,7 +134,8 @@ type WalletKitClient interface {
 	// child-pays-for-parent (CPFP) scenario. If the given output has been
 	// used in a previous BumpFee call, then a transaction replacing the
 	// previous is broadcast, resulting in a replace-by-fee (RBF) scenario.
-	BumpFee(context.Context, wire.OutPoint, chainfee.SatPerKWeight) error
+	BumpFee(context.Context, wire.OutPoint, chainfee.SatPerKWeight,
+		...BumpFeeOption) error
 
 	// ListAccounts retrieves all accounts belonging to the wallet by default.
 	// Optional name and addressType can be provided to filter through all the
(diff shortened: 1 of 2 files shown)


walletkit_client.go
```diff
@@ -737,28 +738,69 @@ func (m *walletKitClient) ListSweepsVerbose(ctx context.Context,
...
+func WithImmediate() BumpFeeOption {
+	return func(r *walletrpc.BumpFeeRequest) {
+		r.Immediate = true
+	}
+}
+
+// WithTargetConf is an option for setting the target_conf of BumpFee. If set,
+// the underlying fee estimator will use the target_conf to estimate the
+// starting fee rate for the fee function. The value of feeRate passed to
+// BumpFee will be ignored.
```
ziggie1984, line 755: The conf-target only reflects this behaviour for the 18.5 release, before the conf target had a different meaning:

https://github.com/lightningnetwork/lnd/pull/9470

walletkit_client.go
```diff
@@ -737,28 +738,69 @@ func (m *walletKitClient) ListSweepsVerbose(ctx context.Context,
...
+func WithTargetConf(targetConf uint32) BumpFeeOption {
+	return func(r *walletrpc.BumpFeeRequest) {
+		r.TargetConf = targetConf
+
+		// We unset sat_per_vbyte, because if it is specified, it is
+		// used instead of target_conf.
+		r.SatPerVbyte = 0
```
ziggie1984, line 762: I think that should not be done here, but its already done in LND that it's not allowed to set both.

walletkit_client.go
```diff
@@ -737,28 +738,69 @@ func (m *walletKitClient) ListSweepsVerbose(ctx context.Context,
...
+	}
+}
+
+// WithBudget is an option for setting the budget of BumpFee. It is the max
+// amount in sats that can be used as the fees. Setting this value greater than
+// the input's value may result in CPFP - one or more wallet utxos will be used
+// to pay the fees specified by the budget. If not set, for new inputs, by
+// default 50% of the input's value will be treated as the budget for fee
+// bumping; for existing inputs, their current budgets will be retained.
```
ziggie1984, line 771: maybe just let the descripiton of the RPC speak for itself and do not require these extensive comments ?

ziggie1984 (changes requested):
Make LND release 18.5 min requirement.

walletkit_client.go
starius: I added a check that LND is at least 0.18.5 only if WithTargetConf is used. This commit can be undone after minimalCompatibleVersion is bumbed to 0.18.5.

walletkit_client.go
starius: I removed this code and added a check that if WithTargetConf is used, feeRate is 0. If both are specified, it is not clear, what the user wants and it is better to fail than to risk doing a wrong thing (e.g. overpay).

walletkit_client.go
starius: I copy-pasted this from https://lightning.engineering/api-docs/api/lnd/wallet-kit/bump-fee/
All parts of this description seem to be important since they define the behavior. IMHO better to keep this in the documentation as well so it is easier to access this information.

starius:
Tested. Works as expected!

walletkit_client.go
```diff
@@ -737,28 +742,92 @@ func (m *walletKitClient) ListSweepsVerbose(ctx context.Context,
...
-			},
-			SatPerByte: uint32(feeRate.FeePerKVByte() / 1000),
-			Force:      false,
+	req := &walletrpc.BumpFeeRequest{
+		Outpoint: &lnrpc.OutPoint{
+			TxidBytes:   op.Hash[:],
+			OutputIndex: op.Index,
 		},
-	)
+		SatPerVbyte: uint64(feeRate.FeePerKVByte() / 1000),
```
ziggie1984, line 803:
```suggestion
SatPerVbyte: uint64(feeRate.FeePerVByte()),
```

walletkit_client.go
```diff
@@ -737,28 +742,92 @@ func (m *walletKitClient) ListSweepsVerbose(ctx context.Context,
 	return result, nil
 }
 
+// BumpFeeOption customizes a BumpFee call.
+type BumpFeeOption func(*walletrpc.BumpFeeRequest)
+
+// WithImmediate is an option for enabling the immediate mode of BumpFee. The
+// sweeper will sweep this input without waiting for the next batch.
```
ziggie1984, line 749: ... without waiting for the next block.

walletkit_client.go
ziggie1984: also specify that since 18.5 if the budget is set, the deadline has to be set too.

ziggie1984 (approved):
LGTM pending new comments.

walletkit_client.go
starius: Applied. Thanks!

walletkit_client.go
starius: Fixed. Thanks!

starius: Does it make sense to fix the description of the field `Immediate` as well?
https://pkg.go.dev/github.com/lightningnetwork/lnd@v0.18.5-beta/lnrpc/walletrpc#BumpFeeRequest

walletkit_client.go
ziggie1984: I was a bit confused, if you select the deadline you need to specify the budget, not the other way around.

https://github.com/lightningnetwork/lnd/blob/68105be1ebbd8cae18c74f112d85382702477d80/lnrpc/walletrpc/walletkit_server.go#L1022-L1023

ziggie1984: yes, it would be cool if you'd open a PR.

starius: In this version there is no deadline setting, since it is not supported by 0.18.4 anyway.
We can add the comment when we add the deadline setting (for lndclient 0.18.5+).

walletkit_client.go
starius: Opened a PR here to fix the description: https://github.com/lightningnetwork/lnd/pull/9518

ziggie1984 (approved):
LGTM

guggero (approved):

guggero:
Pushed a new tag with this: `v0.18.4-11`

PR #223, opened 2025-05-23
walletkit: add  SubmitPackage

bhandras:
This is experimental and is pending on https://github.com/lightningnetwork/lnd/pull/9784

#### Pull Request Checklist


guggero:
BTW, just as an info: Each sub client of the `lndclient` now has a method `RawClientWithMacAuth` which allows you to call any underlying RPC method without needing to add it to the code here first. Could be useful during development.

bhandras:
Closing this as @ellemouton took over in https://github.com/lightninglabs/lndclient/pull/280

PR #224, opened 2025-05-23
closechannel: add functional options

hieblmi:
Adds ~the NoWait flag~ functional options to `CloseChannel`.

### DIFF
--- a/lightning_client.go
+++ b/lightning_client.go
@@ -181,8 +181,9 @@ type LightningClient interface {
 
 	// CloseChannel closes the channel provided.
 	CloseChannel(ctx context.Context, channel *wire.OutPoint,
-		force bool, confTarget int32, deliveryAddr btcutil.Address) (
-		chan CloseChannelUpdate, chan error, error)
+		force bool, confTarget int32, deliveryAddr btcutil.Address,
+		opts ...CloseChannelOption) (chan CloseChannelUpdate,
+		chan error, error)
 
 	// UpdateChanPolicy updates the channel policy for the passed chanPoint.
 	// If the chanPoint is nil, then the policy is applied for all existing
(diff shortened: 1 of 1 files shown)


lightning_client.go
```diff
@@ -2873,7 +2873,7 @@ type PaymentRequest struct {
 	// Value is the value of the payment request in millisatoshis.
 	Value lnwire.MilliSatoshi
 
-	/// Timestamp of the payment request.
+	// / Timestamp of the payment request.
```
bhandras, line 2876: nit: extra space

guggero, line 2876: nit: just remove the third character instead?

lightning_client.go
```diff
@@ -3093,11 +3093,11 @@ func (p *ChannelClosedUpdate) CloseTxid() chainhash.Hash {
 // sending an EOF), we close the updates and error channel to signal that there
 // are no more updates to be sent. It takes an optional delivery address that
 // funds will be paid out to in the case where we cooperative close a channel
-// that *does not* have an upfront shutdown addresss set.
+// that *does not* have an upfront shutdown address set.
 func (s *lightningClient) CloseChannel(ctx context.Context,
```
bhandras, line 3097: You'll also need to update the interface definition in `LightningClient`.

lightning_client.go
```diff
@@ -3093,11 +3093,11 @@ func (p *ChannelClosedUpdate) CloseTxid() chainhash.Hash {
...
 	channel *wire.OutPoint, force bool, confTarget int32,
-	deliveryAddr btcutil.Address) (chan CloseChannelUpdate, chan error,
-	error) {
+	deliveryAddr btcutil.Address, noWait bool) (chan CloseChannelUpdate,
```
guggero, line 3099: To avoid needing changes in all clients that use this method, should we use functional options instead?

Also, the method signature needs to be changed in the `LightningClient` interface.

bhandras (commented):
Looks good, just two nits.

lightning_client.go
hieblmi: fixed

lightning_client.go
hieblmi: fixed.

lightning_client.go
```diff
@@ -3085,6 +3086,33 @@ func (p *ChannelClosedUpdate) CloseTxid() chainhash.Hash {
 	return p.CloseTx
 }
 
+// CloseChannelOption is a functional type for an option that modifies a
+// CloseChannelRequest.
+type CloseChannelOption func(r *lnrpc.CloseChannelRequest)
+
+// SatPerVbyte is an option for setting the fee rate of a CloseChannelRequest.
+func SatPerVbyte(satPerVbyte uint64) CloseChannelOption {
```
bhandras, line 3094: nit: shall we use `chainfee.SatPerVByte` here and for `MaxFeePerVByte`?

lightning_client.go
```diff
@@ -3085,6 +3086,33 @@ func (p *ChannelClosedUpdate) CloseTxid() chainhash.Hash {
...
+// willing to pay on a CloseChannelRequest.
+func MaxFeePerVbyte(maxFeePerVbyte uint64) CloseChannelOption {
+	return func(r *lnrpc.CloseChannelRequest) {
+		r.MaxFeePerVbyte = maxFeePerVbyte
+	}
+}
+
+// WithNoWait is an option for setting the NoWait flag on an
+// CloseChannelRequest.
+func WithNoWait() CloseChannelOption {
```
bhandras, line 3110: If only Go would have namespaces so we don't pollute the global namespace with these generic names. (no action required, just a rant)

bhandras (approved):
LGTM 🎉

lightning_client.go
hieblmi: fixed

PR #233, opened 2025-07-08
RegisterSpendNtfn: support reorg channel

starius:
The corresponding API of LND already supports notifying the caller about a reorg. I reused NotifierOption's used for RegisterConfirmationsNtfn already to pass a channel notified upon a reorg affecting a reported confirmation.

#### Pull Request Checklist

### DIFF
--- a/chainnotifier_client.go
+++ b/chainnotifier_client.go
@@ -71,8 +71,9 @@ type ChainNotifierClient interface {
 		chan error, error)
 
 	RegisterSpendNtfn(ctx context.Context,
-		outpoint *wire.OutPoint, pkScript []byte, heightHint int32) (
-		chan *chainntnfs.SpendDetail, chan error, error)
+		outpoint *wire.OutPoint, pkScript []byte, heightHint int32,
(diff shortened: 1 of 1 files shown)


chainnotifier_client.go
```diff
@@ -162,6 +173,14 @@ func (s *chainNotifierClient) RegisterSpendNtfn(ctx context.Context,
 		return nil
 	}
 
+	processReorg := func() {
+		if opts.ReOrgChan == nil {
+			return
+		}
+
+		opts.ReOrgChan <- struct{}{}
```
guggero, line 181: I think we should also listen on `ctx.Done()` here, see `RegisterConfirmationsNtfn`.

guggero (changes requested):
Thanks for the feature!

chainnotifier_client.go
bhandras: On the same note, I think we should also do this before writing to the `errChan` (and fix conf notifier too).

starius: Added this in another commit.

guggero (approved):

bhandras (approved):
LGTM 🎉

PR #238, opened 2025-08-25
linter: enforce some constraints from code guidelines

bitromortac:
We enable linter caching, enable and fix the long line linter (https://github.com/lightningnetwork/lnd/blob/master/docs/development_guidelines.md#80-character-line-length) as well as add some rules concerning exported symbols (https://github.com/lightningnetwork/lnd/blob/master/docs/development_guidelines.md#code-documentation-and-commenting).

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -26,6 +26,14 @@ linters-settings:
       - G115 # Integer overflow conversion.
   staticcheck:
     checks: ["-SA1019"]
+  revive:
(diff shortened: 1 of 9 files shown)


hieblmi (approved):
Thanks! LGTM!

In a next step we can update `golangci-lint` to `v2.4.0`.

bitromortac:
Thanks for the reviews!

PR #261, opened 2026-01-13
signer: add MuSig2 combined nonce registration methods

Roasbeef:
This commit adds two new methods to the SignerClient interface:

- MuSig2RegisterCombinedNonce: Registers a pre-aggregated combined nonce for a session. This is an alternative to MuSig2RegisterNonces and is used when a coordinator has already aggregated all individual nonces.

- MuSig2GetCombinedNonce: Retrieves the combined nonce for a session after it becomes available.

These methods are needed for coordinator-based MuSig2 signing workflows. Also updates lnd dependency to include sqldb v1.0.12.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,14 +2,14 @@ module github.com/lightninglabs/lndclient
 
 require (
 	github.com/btcsuite/btcd v0.24.3-0.20250318170759-4f4ea81776d6
-	github.com/btcsuite/btcd/btcec/v2 v2.3.4
+	github.com/btcsuite/btcd/btcec/v2 v2.3.6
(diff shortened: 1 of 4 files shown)


ellemouton (approved):
LGTM!

bhandras (commented):
I don't think we should merge this to `master` as it's tracking v0.20, rather we should create `lnd-21-0` branch merge there and tag as `v0.21.0-1`?

Roasbeef:
> I don't think we should merge this to master as it's tracking v0.20, rather we should create lnd-21-0 branch merge there and tag as v0.21.0-1?

Done 👍

bhandras (approved):

PR #262, opened 2026-01-22
lndclient: silence expected close error on shutdown

starius:
During planned termination, conn.Close() can return [ErrClientConnClosing from gRPC package](https://pkg.go.dev/google.golang.org/grpc#pkg-variables). This is expected and should not be logged as an error. Downgrade that case to debug so logs stay clean during shutdown.

Original log line:
```
[ERR] LNDC: Error closing lnd connection: rpc error: code = Canceled desc = grpc: the client connection is closing
```

#### Pull Request Checklist

### DIFF
--- a/lnd_services.go
+++ b/lnd_services.go
@@ -332,7 +332,15 @@ func NewLndServices(cfg *LndServicesConfig) (*GrpcLndServices, error) {
 
 	cleanupConn := func() {
 		closeErr := conn.Close()
-		if closeErr != nil {
+		switch {
+		case closeErr == nil:
+			// No error.
+
+		case errors.Is(closeErr, grpc.ErrClientConnClosing):
+			log.Debugf("LND connection is already closing: %v",
+				closeErr)
+
+		default:
 			log.Errorf("Error closing lnd connection: %v", closeErr)
 		}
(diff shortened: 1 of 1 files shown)


lnd_services.go
```diff
@@ -332,7 +332,10 @@ func NewLndServices(cfg *LndServicesConfig) (*GrpcLndServices, error) {
 
 	cleanupConn := func() {
 		closeErr := conn.Close()
-		if closeErr != nil {
+		if errors.Is(closeErr, grpc.ErrClientConnClosing) {
```
ziggie1984: An optional maybe better version:

```
switch {
case closeErr == nil:
    // No error, do nothing or log a simple success
case errors.Is(closeErr, grpc.ErrClientConnClosing):
    log.Debugf("LND connection is already closing: %v", closeErr)
default:
    log.Errorf("Error closing lnd connection: %v", closeErr)
}
```

ziggie1984 (approved):
LGTM

lnd_services.go
starius: Applied. Thanks!

hieblmi (approved):

PR #264, opened 2026-02-12
go.mod: bump lnd to v0.20.1-beta, go to 1.25.5

hieblmi:


bhandras (approved):

tools/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM golang:1.24.9-bookworm
+FROM golang:1.25-bookworm
```
starius: Should we use 1.26 in tools and CI?

Also, should we pin the patch version here? Like `golang:1.25.5-bookworm` or `golang:1.26.0-bookworm`

go.mod
```diff
@@ -193,4 +193,4 @@ require (
 // allows us to specify that as an option.
 replace google.golang.org/protobuf => github.com/lightninglabs/protobuf-go-hex-display v1.33.0-hex-display
 
-go 1.24.9
+go 1.25.5
```
starius, line 196: CI still pins GO_VERSION: 1.24.6 in .github/workflows/main.yml.

Since Go 1.21, the go command will auto‑download a newer toolchain when go.mod requires a higher version and GOTOOLCHAIN=auto (the default), that is why CI passed:

```
Run actions/setup-go@v5
Setup go version spec ~1.24.6
Found in cache @ /opt/hostedtoolcache/go/1.24.12/x64
Added go to the path
Successfully set up Go version ~1.24.6
go: downloading go1.25.5 (linux/amd64)
/opt/hostedtoolcache/go/1.24.12/x64/bin/go env GOMODCACHE
/opt/hostedtoolcache/go/1.24.12/x64/bin/go env GOCACHE
/home/runner/go/pkg/mod
/home/runner/.cache/go-build
Cache is not found
go version go1.25.5 linux/amd64
```

But ideally we keep them in sync.

I propose to move .github/workflows/main.yml to 1.26. In LND the common practice is to keep go.mod at previous Go version (i.e. 1.25) and CI/toolchain at the latest Go version (i.e. 1.26).

starius (changes requested):

tools/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/lightninglabs/lndclient/tools
 
-go 1.24.6
+go 1.26
```
starius: I think in go.mod it should stay 1.25.5

tools/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM golang:1.24.9-bookworm
+FROM golang:1.26-bookworm
```
starius: `1.26.0-bookworm`

So it is fully pinned.

tools/go.mod
hieblmi: fixed

tools/Dockerfile
hieblmi: fixed

bhandras (approved):

PR #267, opened 2026-03-17
lndclient: add FailureMessage and FailureCode to InterceptedHtlcResponse

calvinrzachman:
## Description

The `routerrpc` HTLC interception API in **_lnd_** already supports `failure_message` and `failure_code` on `ForwardHtlcInterceptResponse`, but **_lndclient_** never exposed them. This PR brings the library in line with the proto by adding `FailureMessage` and `FailureCode` to `InterceptedHtlcResponse` and wiring them through `rpcInterceptorResponse`.

`FailureMessage` accepts a pre-encrypted onion error blob. When set, **_lnd_** relays the bytes through `IntermediateEncrypt` rather than constructing a new error with `EncryptFirstHop`. This allows a virtual node behind the interceptor to produce errors attributed to itself instead of the intercepting node.

`FailureCode` selects a specific failure code for errors that **_lnd_** constructs on behalf of the intercepting node. Without either field, **_lnd_** defaults to `TemporaryChannelFailure`.

#### Pull Request Checklist

### DIFF
--- a/router_client.go
+++ b/router_client.go
@@ -395,6 +395,19 @@ type InterceptedHtlcResponse struct {
 	// intercepted.
 	Action InterceptorAction
 
+	// FailureMessage is a pre-encrypted onion error to return when the
+	// action is fail. When set, lnd relays the bytes using
+	// IntermediateEncrypt, preserving the error attribution of the
+	// original encrypter.
+	FailureMessage []byte
+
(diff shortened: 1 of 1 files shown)


Roasbeef (approved):
LGTM 🫆

bitromortac (approved):
LGTM :zap:

PR #268, opened 2026-04-14
Preserve AddInvoiceData fields across invoice wrappers

starius:
## Summary

`lightningClient.AddInvoice` was silently dropping several fields from `invoicesrpc.AddInvoiceData` when building the outgoing `lnrpc.Invoice` request. That meant callers who set `RouteHints`, `FallbackAddr`, `Amp`, or `BlindedPathCfg` could get an invoice that ignored those settings.

This branch fixes `AddInvoice` to forward all applicable fields on the standard invoice path:

- `RouteHints`
- `FallbackAddr`
- `Amp`
- `BlindedPathCfg` (for the overlapping subset supported by the RPC)

It also cleans up hold-invoice semantics on that path. `Hash` and `HodlInvoice` are hold-invoice-specific inputs, so `lightningClient` now warns when they are passed and no longer forwards `Hash` into `lnrpc.Invoice.RHash`, which lnd documents as output-only.

For `BlindedPathCfg`, only `MinNumPathHops` can be represented through the standard RPC, so it is mapped to `lnrpc.BlindedPathConfig.NumHops`. Other blinded-path tuning fields (`RoutePolicyIncrMultiplier`, `RoutePolicyDecrMultiplier`, `DefaultDummyHopPolicy`) have no counterpart in the RPC proto, so the wrapper logs a warning and lnd falls back to its defaults for those.

As a follow-up for existing asymmetry, `invoicesClient.AddHoldInvoice` now also warns when callers pass `Amp` or `BlindedPathCfg`, since the hold-invoice RPC has no fields for them.

## Test Plan

- Extend `TestLightningClientAddInvoice` with direct regression coverage for: fallback address, AMP, hash/hodl inputs being ignored on the standard invoice path, blinded invoices, blinded invoices with `MinNumPathHops == 0`, and route hints.
- Keep `assertAddInvoiceArgs` for field-by-field verification of the outgoing `lnrpc.Invoice` request.
- Extend `TestInvoiceClientAddInvoiceParity` to compare all shared explicit fields between `AddInvoice` and `AddHoldInvoice`: `Memo`, `ValueMsat`, `DescriptionHash`, `Expiry`, `FallbackAddr`, `CltvExpiry`, `Private`, and `RouteHints`.
- Use path-specific parity fixtures so `Preimage` and `Hash` are no longer set together in the same `AddInvoiceData`.
- Add `TestInvoicesClientAddHoldInvoiceIgnoresUnsupportedFields` to document that `AddHoldInvoice` still forwards the supported request fields unchanged when unsupported `Amp` / `BlindedPathCfg` inputs are present.

#### Pull Request Checklist


hieblmi (approved):
LGTM, the `warnf`s keep the changes backwards compatible!

bhandras (approved):
LGTM, thank you! 🎉

starius:
Pushed tag `v0.20.0-8` with this fix.

PR #269, opened 2026-04-18
lndclient: attach key locator when any component is non-zero

Roasbeef:
In this PR, we fix a latent bug in `marshallSignDescriptors`'s `fullDescriptor` inner closure (the one used by `SignOutputRawKeyLocator`). The helper currently attaches the RPC `KeyLoc` only when both `KeyLocator.Family != 0` *and* `KeyLocator.Index != 0`. That `&&` is wrong: it silently drops the locator for any valid key whose slot has a zero component, and callers that need the locator to re-derive the key (i.e., they can't just scan for a matching pubkey) end up with a descriptor that no longer identifies the key.

The most visible case is LND's node identity key at `keychain.KeyFamilyNodeKey` (family 6) / index 0. Family 6 is meaningful, index 0 is meaningful (it's the first key), so a caller that populates both ends up with a descriptor that forwards neither. The same goes for any first-index key in a non-zero family, which is exactly what a restored-from-seed wallet has before anything else is derived. We hit this downstream in `darepo-client` and had to carry a full local copy of `SignOutputRaw` just to relax this one predicate. Once this lands the local copy can go away.

The fix flips `&&` to `||`: attach the locator whenever either the family or the index is non-zero, and only omit it for the all-zero `KeyLocator` (which is the struct's zero value and therefore indistinguishable from "unset" at call sites). We also rewrite the surrounding comment block so the new rule and its rationale are clear at the call site.

The legacy partial descriptor path used by `SignOutputRaw` is intentionally untouched. Loop and other downstream callers have adjusted themselves to its "populate either the pubkey or the locator, not both" shape, which is the whole reason `SignOutputRawKeyLocator` exists as a separate method in the first place.

## Test Plan

A new `signer_client_test.go` pins the contract with table-driven tests:

- `TestMarshallSignDescriptorsFullLocator` walks the four (family, index) quadrants against the `fullDescriptor` path. The `family-only_locator_is_attached` subtest (family 6, index 0) is the specific regression guard and fails against the old `&&` predicate.
- `TestMarshallSignDescriptorsPartialOnlyOneOf` covers the legacy `partialDescriptor` path to pin that it still populates either the pubkey or the locator but never both, so we don't regress Loop or similar downstream callers.

~~~
$ go test -run '^TestMarshallSignDescriptors' ./...
ok  	github.com/lightninglabs/lndclient	0.334s
~~~

### DIFF
--- a/signer_client_test.go
+++ b/signer_client_test.go
@@ -1,0 +1,203 @@
+package lndclient
+
+import (
+	"testing"
+
+	"github.com/btcsuite/btcd/btcec/v2"
+	"github.com/btcsuite/btcd/wire"
+	"github.com/lightningnetwork/lnd/keychain"
+	"github.com/stretchr/testify/require"
+)
+
+// newTestPubKey returns a fresh public key for use in descriptor
+// marshalling tests.
+func newTestPubKey(t *testing.T) *btcec.PublicKey {
+	t.Helper()
+
+	priv, err := btcec.NewPrivateKey()
+	require.NoError(t, err)
+
+	return priv.PubKey()
+}
+
+// TestMarshallSignDescriptorsFullLocator pins the contract of the full
+// descriptor marshaller used by SignOutputRawKeyLocator: the key locator
+// is attached whenever either the family or the index is non-zero, and
+// only the all-zero KeyLocator (the struct's zero value, indistinguishable
+// from "unset") is omitted.
+//
+// The family-6 / index-0 case is the regression guard. A prior version of
+// the helper required both components to be non-zero and therefore
+// silently dropped the locator for LND's node identity key slot. Signers
+// that must re-derive the key from its locator — restored-from-seed
+// wallets, or family-pinned identity keys — could not resolve the key.
+func TestMarshallSignDescriptorsFullLocator(t *testing.T) {
+	t.Parallel()
+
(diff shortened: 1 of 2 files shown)


signer_client_test.go
```diff
@@ -0,0 +1,179 @@
...
+//
+// The family-6 / index-0 case is the regression guard. A prior version of
+// the helper required both components to be non-zero and therefore
+// silently dropped the locator for LND's node identity key slot. Signers
+// that must re-derive the key from its locator — restored-from-seed
+// wallets, or family-pinned identity keys — could not resolve the key.
+func TestMarshallSignDescriptorsFullLocator(t *testing.T) {
+	t.Parallel()
+
+	pubKey := newTestPubKey(t)
```
starius, line 37: I propose to add two more tests where `KeyDesc.PubKey` is nil and either `family!=0` or `index!=0`. I.e. replicate "family-only locator is attached" and "index-only locator is attached" for the `KeyDesc.PubKey=nil` case.

starius (approved):
**LGTM!** 💾

Great catch! I propose to add a couple of test cases and to remove the redundant `tc := tc` (lint finding).

ziggie1984 (approved):
LGTM, this should not effect any implementations using lndclient with LND.

PR #273, opened 2026-05-04
routerclient: add payment request route fee estimates

hieblmi:
- Add `EstimateRouteFeeWithPaymentRequest` for payment-request probing through LND's `EstimateRouteFee` RPC:
    - payment-request probing via `PaymentRequest`
    - probe timeout via `Timeout`
  - Add `EstimateRouteFeeResponse` so callers can access `RoutingFee`, `TimeLockDelay`, and `FailureReason` for probe estimates.
  - Keep the existing `EstimateRouteFee(ctx, dest, amt)` method as the graph-based route-fee API without deprecating it.
  - Update macaroon recipe method mapping so the new method maps to the existing `/routerrpc.Router/EstimateRouteFee` permission.
  - Add unit tests for graph estimates, payment-request probe estimates, and timeout validation.


starius:
EstimateRouteFeeWithRequest is oneof (old EstimateRouteFee , PaymentRequest + Timeout).

Wouldn't it be better to add another method for the PaymentRequest + Timeout variant and avoid EstimateRouteFee deprecation? Given the new method is split between these two variants anyway.

litbot-9000:
@starius: review reminder

hieblmi:
Thanks @starius, this makes sense. I created
```
estimateRouteFee(ctx context.Context,
	rpcReq *routerrpc.RouteFeeRequest) (*EstimateRouteFeeResponse, error)
```
which is called by old and new method.

router_client.go
```diff
@@ -48,6 +48,12 @@ type RouterClient interface {
 	EstimateRouteFee(ctx context.Context, dest route.Vertex,
 		amt btcutil.Amount) (lnwire.MilliSatoshi, error)
 
+	// EstimateRouteFeeWithPaymentRequest estimates routing costs by probing
+	// with a payment request.
+	EstimateRouteFeeWithPaymentRequest(ctx context.Context,
+		paymentRequest string,
+		timeout time.Duration) (*EstimateRouteFeeResponse, error)
```
starius: A method name ending with "Request" is confusing, because normally it is a request type which follows this naming convention.

Since the main differentiator of this method is an active probing based on this payment request, I propose to rename it:

```go
EstimateRouteFeeWithProbe(ctx context.Context, invoice string,
  timeout time.Duration) (*EstimateRouteFeeWithProbeResponse, error)
```

I also renamed `paymentRequest` to `invoice` which is more common and understandable name. And response type to EstimateRouteFeeWithProbeResponse.

Also update the rename map:
```
"EstimateRouteFeeWithProbe": "EstimateRouteFee"
```
and tests:
```
TestEstimateRouteFeeWithProbe, …WithProbeRejectsInvalidTimeout
```

router_client.go
```diff
@@ -607,18 +629,57 @@ func (r *routerClient) trackPayment(ctx context.Context,
...
-		return 0, err
+		return nil, err
 	}
 
-	return lnwire.MilliSatoshi(rpcRes.RoutingFeeMsat), nil
+	return &EstimateRouteFeeResponse{
+		RoutingFee:    lnwire.MilliSatoshi(rpcRes.RoutingFeeMsat),
+		TimeLockDelay: rpcRes.TimeLockDelay,
+		FailureReason: rpcRes.FailureReason,
+	}, nil
```
starius, line 833: On a probe that fails to reach the destination, lnd returns a successful RPC with a zeroed fee and the reason in-band.

So the idiomatic
```go
resp, err := EstimateRouteFee…()
if err != nil {
   …
} else {
   use(resp.RoutingFee)
}
```
silently reads RoutingFee == 0 and treats an unroutable destination as a free route. The struct field is documented, but the method doc says nothing, so the trap is easy to fall into.

Fix: call it out in the method doc ("a nil error does not imply success - check FailureReason == FAILURE_REASON_NONE before trusting RoutingFee/TimeLockDelay"), and consider whether a non-NONE reason should be surfaced as an error rather than a zero-value struct. Passing it through is a defensible choice, but it must be loud.

router_client.go
```diff
@@ -48,6 +48,12 @@ type RouterClient interface {
 	EstimateRouteFee(ctx context.Context, dest route.Vertex,
 		amt btcutil.Amount) (lnwire.MilliSatoshi, error)
 
+	// EstimateRouteFeeWithPaymentRequest estimates routing costs by probing
+	// with a payment request.
+	EstimateRouteFeeWithPaymentRequest(ctx context.Context,
```
starius: Context can't bound the call: worth one doc line so callers don't rely on ctx deadline/cancel for the probe.

router_client.go
```diff
@@ -607,18 +629,57 @@ func (r *routerClient) trackPayment(ctx context.Context,
...
+	error) {
+
+	if timeout < 0 {
+		return nil, fmt.Errorf("timeout must not be negative")
+	}
+
+	rpcReq := &routerrpc.RouteFeeRequest{
+		PaymentRequest: paymentRequest,
+	}
+	if timeout > 0 {
```
starius, line 806: I propose to document that timeout=0 means LND chooses its default, which is `DefaultPaymentTimeout=60` seconds.

router_client.go
```diff
@@ -48,6 +48,12 @@ type RouterClient interface {
 	EstimateRouteFee(ctx context.Context, dest route.Vertex,
 		amt btcutil.Amount) (lnwire.MilliSatoshi, error)
 
+	// EstimateRouteFeeWithPaymentRequest estimates routing costs by probing
+	// with a payment request.
+	EstimateRouteFeeWithPaymentRequest(ctx context.Context,
+		paymentRequest string,
```
starius: Zero-amount invoices are rejected server-side (router_server.go:530, "amount must be greater than 0") and surface as an error — fine, but undocumented.

router_client.go
starius: A probe that unexpectedly succeeds returns a gRPC error warning that funds are lost (router_server.go:983). Astronomically unlikely (random 256-bit hash); just be aware it arrives as `err != nil`, not in the struct.

router_client_test.go
```diff
@@ -0,0 +1,131 @@
...
+
+func (m *mockRouterRPCClient) EstimateRouteFee(_ context.Context,
+	request *routerrpc.RouteFeeRequest, _ ...grpc.CallOption) (
+	*routerrpc.RouteFeeResponse, error) {
+
+	m.request = request
+	return m.response, m.err
+}
+
+func TestEstimateRouteFeeWithPaymentRequest(t *testing.T) {
```
starius: add godocs in this file

router_client_test.go
```diff
@@ -0,0 +1,131 @@
...
+			TimeLockDelay:  654,
+			FailureReason:  lnrpc.PaymentFailureReason_FAILURE_REASON_NONE,
+		},
+	}
+	client := &routerClient{
+		client: mock,
+	}
+
+	resp, err := client.EstimateRouteFeeWithPaymentRequest(
+		context.Background(), "lnbc1...", 1500*time.Millisecond,
```
starius: use t.Context() here and below

starius (approved):
**LGTM!** 🎉
I have one naming proposal and also found some small issues.

router_client.go
hieblmi: good call, I renamed it.

router_client.go
hieblmi: fixed

PR #275, opened 2026-05-20
Bump lnd to v0.21.0-beta.rc3

ziggie1984:
## Summary
- bump the lnd dependency to the `v0.21.0-beta.rc3` tag
- align the lnd split modules with published tags: `kvdb v1.4.16`, `queue v1.2.0`, and `sqldb v1.0.13`
- update PayInvoice to use routerrpc.SendPaymentV2 after SendPaymentSync removal
- track already completed payments with TrackPaymentV2 so returned fees are not assumed to be zero

## Router RPC tradeoff

PayInvoice remains on the existing LightningClient API, but it now internally uses the router RPC client. This avoids a public API move to RouterClient while matching the supported lnd v0.21.x payment RPC surface. The tradeoff is that PayInvoice now depends on the router subserver and router macaroon in addition to the normal Lightning client wiring.

## Verification
- make build
- make unit
- make unit-race
- make lint


lightning_client.go
```diff
@@ -1564,103 +1572,103 @@ func (s *lightningClient) payInvoice(ctx context.Context, invoice string,
...
-			// TODO: Use structured payment error when available,
-			// instead of this brittle string matching.
-			switch payResp.PaymentError {
-			// Paid successfully.
-			case PaymentResultSuccess:
-				log.Infof(
-					"Payment %v completed", hash,
-				)
+		if status.Code(err) == codes.AlreadyExists {
+			log.Infof("Payment %v already completed", hash)
```
calvinrzachman: nit: The AlreadyExists error also covers `ErrPaymentInFlight`, so "already completed" may not always be correct. Maybe something like "Payment %v already requested/initiated"?

lightning_client.go
```diff
@@ -1564,103 +1572,103 @@ func (s *lightningClient) payInvoice(ctx context.Context, invoice string,
 
 	hash := lntypes.Hash(*payReq.PaymentHash)
 
-	ctx = s.adminMac.WithMacaroonAuth(ctx)
+	ctx = s.routerMac.WithMacaroonAuth(ctx)
 	for {
```
jtobin: Apparently this loop is dead now; the body works as just straight-line code, as there's no more retry case under the streaming API.

jtobin (approved):
LGTM fwiw, though see my one comment re: the 'for' loop in payInvoice now seeming to be dead. I think it should still work fine as-is, though.

calvinrzachman (approved):
Thanks for posting this. I gave it a spin in some of our other projects and it helps clear up this build error:
  ```
  # github.com/lightninglabs/lndclient
  ../../../../pkg/mod/github.com/lightninglabs/lndclient@v0.21.0-rc1/lightning_client.go:1572:17: undefined: lnrpc.SendRequest
  ../../../../pkg/mod/github.com/lightninglabs/lndclient@v0.21.0-rc1/lightning_client.go:1585:28: s.client.SendPaymentSync undefined (type lnrpc.LightningClient has no field or method SendPaymentSync)
  ```
  Just had a one small note. Otherwise this lgtm.

PR #280, opened 2026-06-11
multi: migrate to btcd v2 modules + add WalletKit.SubmitPackage

ellemouton:
Combines the two changes that were split across #280 and #282 into a single
PR for easier review. Two commits:

### 1. `mod: update to btcd v2 modules`

Migrates `lndclient` to the new per-package `v2` modules introduced by `btcd`
`v0.26.0`. The monolithic `btcd` packages (`wire`, `txscript`, `chaincfg`,
`chainhash`, `btcutil`, `psbt`) were split out into their own `/v2` modules,
and the address-handling symbols moved out of `btcutil` into a new `address`
package (imported as `btcaddr` to avoid shadowing local `address` variables).

All dependencies are pinned to released tags — no fork `replace`s:
- `btcd` `v0.26.0` + `btcec/v2` `v2.5.0`
- `btcwallet` `v0.18.0`, `neutrino` `v0.18.0`, `lightning-onion` `v1.4.0`
- `lnd` master (post btcd-v2), `lnd/tlv` `v1.4.0`, `lnd/tor` `v1.2.0`

`btcec/v2` `v2.5.0` requires Go 1.25, so the Go version is bumped to 1.25.11.

### 2. `walletkit: add SubmitPackage client method`

Adds `WalletKitClient.SubmitPackage`, wrapping lnd's new `WalletKit.SubmitPackage`
RPC ([lightningnetwork/lnd#10900](https://github.com/lightningnetwork/lnd/pull/10900), merged).
It serializes the topologically-sorted transactions, forwards an optional
per-transaction max fee rate (`nil` uses the node default), and maps the proto
response into an lndclient-native `SubmitPackageResult` so callers do not depend
on `btcjson`'s submitpackage types.

This lets a zero-fee v3/TRUC parent be relayed via its fee-paying CPFP child
(bitcoind's `submitpackage`) without the caller needing a separate connection
to the chain backend.


walletkit_client.go
```diff
@@ -525,6 +535,72 @@ func (m *walletKitClient) PublishTransaction(ctx context.Context,
 	return err
 }
 
+func (m *walletKitClient) SubmitPackage(ctx context.Context,
+	txns []*wire.MsgTx,
+	maxFeeRate *float64) (*btcjson.SubmitPackageResult, error) {
```
Roasbeef: I don't think we should leak `btcjson` here (notice how it was just imported with this diff). Instead we should make a higher level wrapper type like we do elsewhere for related APIs.

Roasbeef (changes requested):

walletkit_client.go
ellemouton: Addressed — the method no longer leaks `btcjson`. It now maps the proto response into lndclient-native `SubmitPackageResult` / `SubmitPackageTxResult` types (with `chainhash.Hash` txids), so callers don't depend on `btcjson`'s submitpackage types. The `btcjson` import is gone from the file.

GeorgeTsagk (approved):
LGTM https://github.com/lightninglabs/lndclient/pull/280/changes/ee76ab9ecc2c8b3d4765727966c26b46a17ba2e5

tACK with https://github.com/lightninglabs/taproot-assets/pull/2186

Roasbeef (approved):
LGTM 🌪️

PR #285, opened 2026-08-11
lndclient: fix decoded payment request expiry

starius:
LND returns DecodePayReq expiry as a duration from the invoice timestamp. Add that duration instead of treating it as an absolute Unix timestamp.

See https://github.com/lightningnetwork/lnd/pull/11058

Also renamed `PaymentRequest.Expiry` to `PaymentRequest.ExpiresAt` to prevent existing callers from silently interpreting the changed value semantics. **This is a breaking API change!**

#### Pull Request Checklist

### DIFF
--- a/lightning_client.go
+++ b/lightning_client.go
@@ -2908,8 +2908,8 @@ type PaymentRequest struct {
 	// Timestamp of the payment request.
 	Timestamp time.Time
 
-	// Expiry is the time at which the payment request expires.
-	Expiry time.Time
+	// ExpiresAt is the time at which the payment request expires.
+	ExpiresAt time.Time
 
 	// Description is a description attached to the payment request.
 	Description string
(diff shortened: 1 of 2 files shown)


lightning_client.go
```diff
@@ -2961,7 +2961,11 @@ func (s *lightningClient) DecodePaymentRequest(ctx context.Context,
 	}
 
 	if resp.Expiry != 0 {
-		paymentReq.Expiry = time.Unix(resp.Expiry, 0)
+		// LND reports expiry as a duration from the invoice timestamp.
+		invoiceTimestamp := time.Unix(resp.Timestamp, 0)
+		paymentReq.Expiry = invoiceTimestamp.Add(
+			time.Duration(resp.Expiry) * time.Second,
+		)
 	}
```
GustavoStingelin, line 2969: Since this changes the meaning of the value stored in this field, it is already a behavioral breaking change for existing consumers. What do you think about renaming `Expiry` to `ExpiresAt`? 
The new name would make the absolute-time semantics explicit and cause a compile-time error for existing callers, rather than silently changing the value’s meaning. I also think this breaking change should be noted in the release notes. 

That said, if we can confirm that no one is currently using this field, I’m fine with keeping it as is.

GustavoStingelin: it can also help to avoid misleading, as the 'expiry' for lnd and bolt means duration and not timestamp

starius: Good catched! I renamed it in a separate commit.

GustavoStingelin (approved):
LGTM!

starius:
Tagged as `v0.21.0-3`

